PR #4, opened 2020-11-18
ADR-18: Content mappings flow for explorer and builder in-world

BrianAmadori:


docs/ADR-18-content-mappings-flow-for-explorer-and-builder-in-world.md
```diff
@@ -0,0 +1,262 @@
...
+Kernel->Renderer: create any sdk component\nwith src payload
+Renderer->Renderer: Use baseUrl and mappings\nto solve full asset url.
+```
+!-->
+
+This approach works fine for most of the cases. However, some issues can be observed:
+
+1) For big scenes, a pretty big content mappings payload have to be sent over to Renderer. This can produce performance hiccups.
+
+2) For content coming from dynamic sources (i.e.: builder or builder in world) the chance of hash collisions in the local asset path is high. Any artist can inadvertently name their assets using generic naming, and in this case a collision is going to occur.
```
nchamo, line 79: I would try rephrase (2) and (3) to make it clearer that:
* The main problem with the Builder In-World is that it adds assets dinamically, so our rigid mapping logic doesn't work anymore. This is the same, whether we support custom user assets or not
* The collision is in the local asset path, and not the hash itself. 
* When we support custom user assets, the chances for local asset path collisions becomes really high

docs/ADR-18-content-mappings-flow-for-explorer-and-builder-in-world.md
```diff
@@ -0,0 +1,262 @@
...
+
+* Have an asset catalog Kernel<>Renderer messaging pipeline for adding and removing assets lazily as needed.
+* Make all our components receive `assetId` references that are going to be mapped to this catalog to retrieve the proper `ContentProvider` and solve our asset urls.
+* Our `LoadParcelScenes` flow will no longer provide the mappings. All the asset ids management responsibilities are  going to come from the asset catalog flow.
+
+This is arguably the best solution in terms of solving (1) (2) and (3). However, the refactor cost is high. We would have to adapt all the components to this new approach and re-design the LoadParcelScenes message.
+
+We can split this approach in different stages.
+
+#### Stage 1: MVP for builder in-world
```
nchamo, line 205: Lets make it clearer that this will only be for the Builder In-World feature. The change to the `LoadParcelScenes` should be backwards compatible, to continue to support scenes deployed right now

nchamo (approved):
Great work!

docs/ADR-18-content-mappings-flow-for-explorer-and-builder-in-world.md
BrianAmadori: Edited to improve clarity on those points. Thanks!

docs/ADR-18-content-mappings-flow-for-explorer-and-builder-in-world.md
BrianAmadori: Added some details to add emphasis on backwards compatibility.

PR #10, opened 2021-06-02
feat: add adr 26 - Signup screen to unity

AjimenezDCL:


docs/ADR-26-port-signup-screen-to-unity-.md
```diff
@@ -0,0 +1,47 @@
...
+6) _Website_ dispatches an event with the passport info (_name_ and _email_).
+7) _Kernel_ finishes the signup with the passport info and converges to the normal signin workflow.
+
+## Approach
+
+The _passport_ and _ToS_ screens will be ported into Unity. For _kernel_ the workflow will be almost identical while all the related code for _website_ will be removed.
+
+The new flow should go as this:
+
+1) _Kernel_ requests _Avatar Editor_ in _Unity_.
```
BrianAmadori, line 32: I know that this is maybe a matter of wording. But this sentence shouldn't look like this?
> _Kernel_ requests _start sign-up flow_ to Unity

As this is what is actually happening, as opposed to just opening the _Avatar Editor_.

AjimenezDCL: I wanted to make implicit that the Avatar Editor is brought up as a request from kernel. In fact there's no such a thing as signup process in Unity, just a call to open the avatar editor:

`kernel/packages/shared/session/sagas.ts`
image

AjimenezDCL: In any case I completly agree the wording is strange.

menduz (approved):

PR #25, opened 2021-11-08
feat(ADR-44): add signedFetch

2fd:


docs/ADR-43-signed-fetch.md
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Payload
+ */
+const payload = (
+  [
+    METHOD,
+    PATH,
+    TIMESTAMP,
+    METADATA
```
menduz, line 89: Can we take the chance, and use this ADR to review the sign algorithm?

It seems like the current approach is vulnerable to at least four attacks:
- Reuse signature with different `body` content
- Reuse signature with different `queryString`
- Reuse signature with different `headers`
- Reuse signature with different `origin` or `host`

I'd like to propose the following changes to the signed payload:

1. Prepend the version of the signer algorithm, i.e. `sig1`
2. Append the `queryString` to the signature list (payload in this algorithm)
3. Append the sha256 of the body payload if present, otherwise append empty string
4. Append a list of signed headers and their value
5. Replace the `.join(':')` by `.join('\n')` to account for headers and JSON encoded new lines `\\n`
6. Remove the `.toLowerCase()` of the result since it lowers the security of the algorithm

All of the suggestions take into account what the AwsSignatureAlgorithmV4 does https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

I think we could gather in a call to evaluate the best way to do this to preserve backwards compatibility and make all the changes only once.

menduz (changes requested):

docs/ADR-43-signed-fetch.md
menduz: I'd also suggest to build on giant'shoulders and use the same encoding as they do, since it is battle tested. We can leverage some libraries to do so like https://github.com/RBrNx/agnostic-aws-signature/blob/master/src/index.js or any of the following https://www.npmjs.com/search?q=signature%20v4

docs/ADR-43-signed-fetch.md
```diff
@@ -0,0 +1,131 @@
...
+ * @example
+ * - `/ping`
+ * - `new URL('https://decentraland.org/ping').pathname`
+ */
+const PATH: string = `/ping`
+
+/**
+ * request timestamp
+ * each service decide how long will consider valid a request using this timestamp,
+ * if timestamp is greater than now the request should fail
```
cazala, line 69: should the services decide on when the signature is too old, or should we use an expiration date signed by the user? to me the later sounds more transparent.

docs/ADR-43-signed-fetch.md
menduz: Somebody took the initiative seriously https://httpwg.org/http-extensions/draft-ietf-httpbis-message-signatures.html

PR #30, opened 2021-11-26
feat: lock collection adr

nicosantangelo:
Closes https://github.com/decentraland/builder-server/issues/361

### DIFF
--- a/docs/ADR-48-locking-collections-in-the-builder.md
+++ b/docs/ADR-48-locking-collections-in-the-builder.md
@@ -12,7 +12,7 @@ Publications vary for each type described above:
 
 ## Problem
 
-Publishing is done by sending a meta-transaction. This means that the user will send a transaction and send it to our server where it will ultimately be sent to the corresponding Polygon network to be processed.
+Publishing is done by sending a transaction. This means that the user will craft a transaction and send it to our server where it will ultimately be sent to the corresponding Polygon network to be processed.
 This process is usually pretty quick, but it can still take a while. It's a combination of a remote request to our server, and waiting for a transaction to be confirmed. The problem is therefore, that while we're waiting for the transaction to complete we should **lock** the collection/items that are being published, to avoid having missmatching data between the blockchain and our database.
 
 ### Solution
@@ -21,7 +21,7 @@ We propose a locking mechanism for both types of Collections and Items. We'll lo
 
 To do this we'll have to put two mechanisms in place, for both Decentraland and Third Party data:
 
-- **Decentraland**: Once the user confirms the signature to send the publication, we'll send a POST to the server locking the collection (`/collections/COLLECTION_ID/lock`). This will set the current timestamp (`Date.now()`) on `lock: timestamp` property of the Collection. This will be then checked in the server when returning each Decentraland Collection, adding a `isLocked: boolean` property to each object if the publication is still pending and a day has not gone by:
+- **Decentraland**: Once the user confirms the signature to send the publication, we'll send a POST to the server locking the collection (`/collections/COLLECTION_ID/lock`). This will set the current timestamp (`Date.now()`) on the `lock: timestamp` property of the Collection. This will be then checked in the server when returning each Decentraland Collection, adding a `isLocked: boolean` property to each object if the publication is still pending and a day has not gone by:
 
 ```ts
 function isLocked(collection: Collection) {
@@ -29,6 +29,7 @@ function isLocked(collection: Collection) {
     return false;
   }
 
+  const { lock } = collection;
   const deadline = new Date(lock);
   deadline.setDate(deadline.getDate() + 1);
 
@@ -36,18 +37,20 @@ function isLocked(collection: Collection) {
 }
 ```
 
-- **Third Party**: Third party collections work in a similar way but a key difference outlined before, we publish items not collections. Still the process works the same way. After the user signs the transaction that will lock the items, we send a POST to the server (`/collections/COLLECTION_ID/lock`) locking each the Thrid Party Collection. This will set the current timestamp (`Date.now()`) on `lock: timestamp` property of each the Collection. Up until this point, things are mostly the same, but when the server returns the data it will now check the lock as before and if the lock is smaller than the last item created at date on the graph, which is the same as saying "last publish date":
+- **Third Party**: Third party collections work in a similar way but a key difference outlined before, we publish items not collections.
+  After the user signs the transaction that will lock the items, we send a POST to the server (`/items/lock`) adding each item id to lock to the body of the request. This will set the current timestamp (`Date.now()`) on the `lock: timestamp` property of each Third party Item. When the server returns the data it will add an `isLocked: boolean` property to each item checking if the `lock` is smaller than the last item's `updated_at` date on the graph, which is the same as saying "last publication date", and if day has not gone by:
 
 ```ts
-function isLocked(collection: Collection, lastItem: Item) {
+function isLocked(item: Collection, remoteItem: Item) {
+  const { lock } = item
   const deadline = new Date(lock)
   deadline.setDate(deadline.getDate() + 1)
 
-  return deadline.getTime() > Date.now()  || collection.lock <= lastItem.created_at)
+  return deadline.getTime() > Date.now()  || lock <= remoteItem.updated_at)
 }
 ```
 
-On the front-end, we'll check for this property and disable all components that allow changes that would create a desync, like changing the name of a Decentraland Collection for example.
+Anyone consuming this data can read this property and disable all components that allow changes that would create a desync, like changing the name of a Decentraland Collection for example.
 
 ## Implementation
 
--- a/docs/ADR-48-locking-collections-in-the-builder.md
+++ b/docs/ADR-48-locking-collections-in-the-builder.md
@@ -1,0 +1,63 @@
+# Locking Collections in the builder
+
+## Abstract
+
+The builder allows users to create items which are composed of GLB or PNG files, and group them in Collections. There's a distinction here between Decentraland Collections and Third Party Collections, we'll point out the differences as they arise.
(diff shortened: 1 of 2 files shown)


docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -0,0 +1,63 @@
...
+Managing items and collections is done off-chain until you decide to publish. This means that you can take items in and out of collections, modify their data, etc and it will be saved in our database. Once the publication occurs, some of the data (like the Decentraland's collection name for example) cannot change, and will be stored both in the blockchain and in our database
+
+Publications vary for each type described above:
+
+- Decentraland: You publish the entire collection with all it's items
+- Third Party: You publish each item individually, but you can do it in batches
+
+## Problem
+
+Publishing is done by sending a meta-transaction. This means that the user will send a transaction and send it to our server where it will ultimately be sent to the corresponding Polygon network to be processed.
```
nachomazzara, line 15: I may say `sending a transaction`. It is the same if it is a meta or normal tx

docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const deadline = new Date(lock);
+  deadline.setDate(deadline.getDate() + 1);
+
+  return deadline.getTime() > Date.now();
+}
+```
+
+- **Third Party**: Third party collections work in a similar way but a key difference outlined before, we publish items not collections. Still the process works the same way. After the user signs the transaction that will lock the items, we send a POST to the server (`/collections/COLLECTION_ID/lock`) locking each the Thrid Party Collection. This will set the current timestamp (`Date.now()`) on `lock: timestamp` property of each the Collection. Up until this point, things are mostly the same, but when the server returns the data it will now check the lock as before and if the lock is smaller than the last item created at date on the graph, which is the same as saying "last publish date":
```
nachomazzara, line 39: How are we preventing items that were not published but created before publishing some of them? Like I have 10 items but only want to publish 4 for the time being.

nachomazzara (changes requested):
Just small comments

docs/ADR-48-locking-collections-in-the-builder.md
LautaroPetaccio: If we're talking about meta-transactions, I would use craft here instead of repeating `send`.
```suggestion
Publishing is done by sending a meta-transaction. This means that the user will craft a transaction and send it to our server where it will ultimately be sent to the corresponding Polygon network to be processed.
```

docs/ADR-48-locking-collections-in-the-builder.md
LautaroPetaccio:
```suggestion
- **Third Party**: Third party collections work in a similar way but a key difference outlined before, we publish items not collections. Still the process works the same way. After the user signs the transaction that will lock the items, we send a POST to the server (`/collections/COLLECTION_ID/lock`) locking the Thrid Party Collection. This will set the current timestamp (`Date.now()`) on `lock: timestamp` property of each the Collection. Up until this point, things are mostly the same, but when the server returns the data it will now check the lock as before and if the lock is smaller than the last item created at date on the graph, which is the same as saying "last publish date":
```

docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -0,0 +1,63 @@
...
+```ts
+function isLocked(collection: Collection, lastItem: Item) {
+  const deadline = new Date(lock)
+  deadline.setDate(deadline.getDate() + 1)
+
+  return deadline.getTime() > Date.now()  || collection.lock <= lastItem.created_at)
+}
+```
+
+On the front-end, we'll check for this property and disable all components that allow changes that would create a desync, like changing the name of a Decentraland Collection for example.
```
LautaroPetaccio, line 50: Are we going to include in this ADR something about how the UI will behave? Like, are we going to let the users reload the page or are we going to ask the server continuously for information about a collection lock?

docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+- **Third Party**: Third party collections work in a similar way but a key difference outlined before, we publish items not collections. Still the process works the same way. After the user signs the transaction that will lock the items, we send a POST to the server (`/collections/COLLECTION_ID/lock`) locking each the Thrid Party Collection. This will set the current timestamp (`Date.now()`) on `lock: timestamp` property of each the Collection. Up until this point, things are mostly the same, but when the server returns the data it will now check the lock as before and if the lock is smaller than the last item created at date on the graph, which is the same as saying "last publish date":
+
+```ts
+function isLocked(collection: Collection, lastItem: Item) {
+  const deadline = new Date(lock)
+  deadline.setDate(deadline.getDate() + 1)
+
+  return deadline.getTime() > Date.now()  || collection.lock <= lastItem.created_at)
```
LautaroPetaccio, line 46: Should we clarify how are we going to get the last item date? Like, we're going to query the graph for the last time an item was published for that collection.

docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -0,0 +1,66 @@
...
+  const { lock } = collection;
+  const deadline = new Date(lock);
+  deadline.setDate(deadline.getDate() + 1);
+
+  return deadline.getTime() > Date.now();
+}
+```
+
+- **Third Party**: Third party collections work in a similar way but a key difference outlined before, we publish items not collections.
+  After the user signs the transaction that will lock the items, we send a POST to the server (`/items/lock`) adding each item id to lock to the body of the request. This will set the current timestamp (`Date.now()`) on the `lock: timestamp` property of each Third party Item. When the server returns the data it will add an `isLocked: boolean` property to each item checking if the `lock` is smaller than the last item's `updated_at` date on the graph, which is the same as saying "last publication date", and if day has not gone by:
```
LautaroPetaccio, line 41:
```suggestion
  After the user signs the transaction that will lock the items, we send a POST to the server (`/items/lock`) with  each item id to lock int the body of the request. This will set the current timestamp (`Date.now()`) on the `lock: timestamp` property of each Third party Item. When the server returns the data it will add an `isLocked: boolean` property to each item that is computed by checking if the `lock` is smaller than the last item's `updated_at` date on the graph, which is the same as saying "last publication date", and if day has not gone by:
```

docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -0,0 +1,66 @@
...
+
+  return deadline.getTime() > Date.now();
+}
+```
+
+- **Third Party**: Third party collections work in a similar way but a key difference outlined before, we publish items not collections.
+  After the user signs the transaction that will lock the items, we send a POST to the server (`/items/lock`) with each item id to lock in the body of the request. This will set the current timestamp (`Date.now()`) on the `lock: timestamp` property of each Third party Item. When the server returns the data it will add an `isLocked: boolean` property to each item that is computed by checking if the `lock` is smaller than the last item's `updated_at` date on the graph, which is the same as saying "last publication date", and if day has not gone by:
+
+```ts
+function isLocked(item: Collection, remoteItem: Item) {
```
nachomazzara, line 44:
```suggestion
function isLocked(item: Item, remoteItem: Item) {
```

nachomazzara: Also, if the item is published, we should `return false`

nachomazzara (changes requested):
I think that we may tell that tp collections can check the lock at the collection level.  If the collection is lock, the URN can't be changed.

docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -42,7 +42,29 @@ function isLocked(element: Collection | Item) {
...
+    - price
+    - beneficiary
+    - metadata
+      - type
+      - name
+      - description
+      - data (the BodyShape and Category)
+
+- **Third Party**: Once the items are published, the URN cannot change anymore. That means that item `token_id`s and virtual collection `urn_suffix`es. For reference, and URN looks like this: `urn:decentraland:{network}:collections-thirdparty:{third-party-name}:{collection-id}:{item-id}` where
+  - network: the network the item is in, `mumbai`, `polygon`, etc
```
nachomazzara, line 64: `matic` instead of `polygon` right?

nachomazzara (changes requested):

docs/ADR-48-locking-collections-in-the-builder.md
nicosantangelo: Shoot, right, I always make a mistake with [common schema's name](https://github.com/decentraland/common-schemas/blob/main/src/dapps/chain-id.ts#L63)

docs/ADR-48-locking-collections-in-the-builder.md
```diff
@@ -42,7 +42,29 @@ function isLocked(element: Collection | Item) {
...
+    - rarity
+    - price
+    - beneficiary
+    - metadata
+      - type
+      - name
+      - description
+      - data (the BodyShape and Category)
+
+- **Third Party**: Once the items are published, the URN cannot change anymore. That means that item `token_id`s and virtual collection `urn_suffix`es. For reference, and URN looks like this: `urn:decentraland:{network}:collections-thirdparty:{third-party-name}:{collection-id}:{item-id}` where
```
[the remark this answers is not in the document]
nicosantangelo: I added it on each section, like this:

  - collection-id: the virtual collection's `urn_suffix`, *used as part of the full URN and locked after the first item is published as it will be used for all subsequent items*
  - item-id: the item's `token_id`, *editable up until publication*
  
you mean adding it here in the first paragraph too? (not against it, just asking)

nachomazzara (approved):

PR #33, opened 2022-01-04
feat:  add ADR 51 - catalyst content validations

guidota:


docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,65 @@
+# Catalyst Conent validations
+
+## Context and Problem Statement
+
+Catalysts receive entity deployments that will be stored and distributed among them. Some validations are executed for each entity deployment, depending on the context and entity type.
+
+Most validations need information about the Catalyst content state and external dependencies like The Graph, which makes them very hard to predict, i.e. we can't ensure an entity deployment will be correct without validating it in given context.
```
agusaldasoro, line 7:
```suggestion
Most validations need information about the Catalyst content state and external dependencies like The Graph, which makes them very hard to predict, i.e. we can't ensure an entity deployment will be correct without validating it in a given context.
```

agusaldasoro (approved):
Great work!! 👏

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,65 @@
...
+
+In the other hand, we have some stateless validations that may be useful to run before actually doing a deployment, like metadata schema and hashes validations.
```
pentreathm, line 9: On the other hand

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,65 @@
...
+
+All validations will know this context and will decide if must do something different, but no validation list per context will exist, i.e. all validations will run on each deployment.
+
+Also, we will get rid of the entity version idea, moving to a timestamp based decision making on validations, so we can avoid duplicated/deprecation logic.
+
+We will execute these validations for all entity types:
+
+- RATE LIMIT
+- NO NEWER
+- RECENT
```
menduz, line 31: I think we should move the concept of "DeploymentContext" completely to the implementation of the server. Yielding RATE_LIMIT, NO_NEWER, and RECENT outside of this spec, since those are implementation details of the `POST /entitites` endpoint, and not really consensus validations for the synchronization protocol.

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+
+On the other hand, we have some stateless validations that may be useful to run before actually doing a deployment, like metadata schema and hashes validations.
+
+Adding an entity type to a Catalyst requires a tedious effort which envolves defining schemas, relations and validations in different repositories.
+
+For instance, adding a validation to the Catalyst can be overwhelming for a developer from outside the Platform team.
+
+## Decision
+
+We will move all validations to a library that will not depend on any other than definitions like schemas and types.
```
menduz, line 17:
```suggestion
Validations will be moved to a library that will not depend on any other than definitions like schemas and types.
```

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+```
+
+- IPFS HASHING
+
+  Validates that all hashes used by the entity are actually IPFS hashes
+
+```
+parameter ADR_45_TIMESTAMP = 1648771200
+
+1. If deployment.timestamp <= ADR_45_TIMESTAMP return
```
menduz, line 57: I think the IPFS_HASH_V0 validation is missing here

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+  Validates that entity pointers are present and not repeated
+
+```
+1. Fail if entity.pointers.length <= 0
+
+2. Fail if Set(entity.pointers).size != entity.pointers.length
+```
+
+- METADATA SCHEMA
+  1
```
menduz, line 76:
```suggestion

```

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+  Validates that uploaded and reported hashes are corrects and files corresponds to snapshots
+
+```
+parameter ADR_45_TIMESTAMP = 1648771200
+parameter ADR_X_SNAPSHOT_FILES = {"face.png", "body.png"}
+
+----
+
+1. For each content file item C in the deployment
+  a. Fail if C is not stored in disk or is not one of the uploaded files
```
menduz, line 104:
```suggestion
  a. Fail if C is not stored in disk or C is not present in the uploaded files
```
What is uploaded files? entity.files?

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+- SIZE
+
+  Validates that the full request size is within limits
+
+```
+parameter ADR_45_TIMESTAMP = 1648771200
+
+----
+
+1. Let MAX_SIZE be the maximum size in bytes allowed by ENTITY_TYPE
```
menduz, line 123: Where is the MAX_SIZE for every ENTITY_TYPE defined?

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+3. Fail validation if CURRENT_SIZE > MAX_SIZE
+```
+
+- ACCESS
+
+  Validates that the pointers are valid, and that the Ethereum address has write access to them
+
+```
+parameter LEGACY_CONTENT_MIGRATION_TIMESTAMP
+parameter DECENTRALAND_ADDRESS
```
menduz, line 138: please specify parameters

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+
+----
+
+1. If deplyoment.timestamp < LEGACY_CONTENT_MIGRATION_TIMESTAMP and deployment.address != DECENTRALAND_ADDRESS
+  return
+
+2. Execute entity access validation
```
menduz, line 145: I think this validation never fails

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+5. Fail if P != deployment.address.toLowerCase()
+```
+
+SCENE entities will have following validations:
+
+- SCENE ACCESS
+
+```
+parameter DECENTRALAND_ADDRESS
+parameter SCENE_LOOKBACK_TIME
```
menduz, line 176: Please specify

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+
+8. Fail if THUMBNAIL_IMAGE height != THUMBNAIL_SIZE
+```
+
+- SIZE
+  Validates wearable files size, excluding thumbnail, is less than expected
+
+```
+parameter ADR_45_TIMESTAMP = 1648771200
+parameter MAX_SIZE = 2
```
menduz, line 305: What is 2? MB?

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+    . Fail if P.network ∈ L2_NETWORKS
+    . Fail if deployment.address != DECENTRALAND_ADDRESS
+```
+
+- THUMBNAIL
+  Validates that given wearable deployment includes a thumbnail with valid format and size
+
+```
+parameter ADR_45_TIMESTAMP = 1648771200
+parameter THUMBNAIL_SIZE = 1024
```
menduz, line 279: What is size? Pixels?

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Status
+
+Accepted.
+
+## Consequences
+
+Will create a starting point to track entities changes.
+
+We will need to create an ADR in order to change any of these statements or add a new entity type, describing which validations will run.
```
menduz, line 330:
```suggestion
An ADR will be created to change any of these statements or add a new entity type, describing which validations will run.
```

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+
+Adding validations will be easier for external team developers.
```
menduz, line 332:
```suggestion
Adding validations will be easier any developer.
```

menduz (changes requested):
Some changes in wording. I'd move the definition of the state of the validations to a separated "live" file that can be mutated on each validation change, by doing that, people can leverage "git blame" to evaluate the diff between each validation and also making easier the PR validations

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+```
+parameter ADR_45_TIMESTAMP = 1648771200
+
+----
+
+1. If deployment.timestamp <= ADR_45_TIMESTAMP return
+
+2. Let entity be the uploaded deployment entity
+
+2. Let SCHEMA be the json schema of current entity.type
```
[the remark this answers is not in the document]
guidota: They are not versioned at the moment. We should implement it or use timestamps for breaking changes.

Schemas by entity type are defined in [`catalyst-commons`](https://github.com/decentraland/catalyst-commons/blob/master/src/entities/entities.ts)

docs/ADR-51-catalyst-content-validations.md
guidota: Yes, uploaded files != content files (some may reference already uploaded contents)

docs/ADR-51-catalyst-content-validations.md
guidota: Here in [`catalyst-commons`](https://github.com/decentraland/catalyst-commons/blob/master/src/entities/entities.ts)

docs/ADR-51-catalyst-content-validations.md
guidota: Yes, will add that assumption

docs/ADR-51-catalyst-content-validations.md
guidota: Yes, will add that assumption

docs/ADR-51-catalyst-content-validations.md
```diff
@@ -0,0 +1,336 @@
...
+
+For instance, adding a validation to the Catalyst can be overwhelming for a developer from outside the Platform team.
+
+## Decision
+
+We will move all validations to a library that will not depend on any other than definitions like schemas and types.
+
+It will expose an interface that will be implemented and required for validations execution, at the moment it means the Catalyst will provide this implementation but any other could do it.
+
+We will redefine the deployment contexts and only take into account: 'LOCAL' and 'SYNCED'.
```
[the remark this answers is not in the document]
guidota: Not needed. Removed.

docs/ADR-51-catalyst-content-validations.md
guidota: We were talking about using only V1 after this time, is that incorrect?

docs/ADR-51-catalyst-content-validations.md
guidota: It's kind of an entry point which will execute the corresponding validation for each entity type. I added more details to the second step.

docs/ADR-51-catalyst-content-validations.md
guidota: Added current max sizes for each entity type.

docs/resources/ADR-51/size.md
```diff
@@ -4,7 +4,11 @@ Validates that the full request size is within limits
 
 ```
 parameter ADR_45_TIMESTAMP = 1648771200
-
+parameter MAX_SIZES_BY_ENTITY_TYPE:
+  - SCENE: 15MB
```
menduz, line 8: Do scenes have variable limits on the size? The main plaza weights around 200MB. It also has a hundred parcels. Does the validation takes that into account?

guidota: Added clarification that the size is per pointer (parcel in scenes).

menduz (approved):

PR #34, opened 2022-01-05
create adr 52: new sync

agusaldasoro:


docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+When there were few deployments all of this worked fine, but when we reached 2M deployments and 50k deployments daily, the time spended retrieving all the history from the other servers and deploying that locally was always bigger than the new amount of deployments. So if you wanted to create a new Catalyst and sync it from scratch it could take you a month and you will never be synced.
+
+
+## Decision
+
+### New Sync
+To build the new sync we put the focus on making the Catalyst work, so we left behind the concept of having all the historical deployments everywhere. This doesn’t mean that we lost that data, it is backed up in s3. But to implement the sync we only took into account the deployed entities that are active.
+
+For the new sync we continue communicating all the catalysts between them, but in a more efficient way:
+1. We decreased the amount of requests for bootstrapping from 400 to 1.
```
menduz, line 48:
```suggestion
1. We decreased the amount of requests for bootstrapping from N to 1. Being `N = TOTAL_DEPLOYMENTS / PAGE_SIZE (paginated list) + TOTAL_DEPLOYMENTS * 1 (audit data)`
```

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+2. Deprecate the endpoint /deployments (which was the most expensive in the db)
```
menduz, line 49:
```suggestion
2. Deprecate the endpoint /deployments (which was the most expensive in the db).
3. Add missing information to /snapshot endpoints to include `authChain` to reduce by TOTAL_DEPLOYMENTS the amount of requests.
```

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+}
+
+while(true) {
+  for (server of allServers) {
+    // GET /pointer-changes : always only the new deployments
+    allDeployments = await server.getNewDeployments() 
+    deployLocally(allDeployments)
+    server.updateLastSeenTimestamp() // the biggest timestamp seen in pointer-changes
+  }
+}
```
menduz, line 70: I think this explains better the concurrency of the synchronization against servers

```suggestion

async function sync(remoteServer) {
  // Bootstrapping
  
  // GET /snapshots : get all entities that are active
  allDeployments = await remoteServer.getSnapshots() 
  deployLocally(allDeployments)
  remoteServer.updateLastSeenTimestamp() // the biggest timestamp seen in snapshots

  // Syncing
  while(remoteServer.isOnline()) {
    // GET /pointer-changes : always only the new deployments
    allDeployments = await remoteServer.getNewDeployments() 
    await deployLocally(allDeployments)
    remoteServer.updateLastSeenTimestamp() // the biggest timestamp seen in pointer-changes
  }
}

for (server of allServers) {
  // concurrently
  sync(server)
}
```

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+In pseudo code both implementations look similar, and they are. But the key here was to stop using `/deployments` endpoint and implement a new performant `/snapshots`.
+
+Why is `/deployments` endpoint so expensive? It’s an endpoint with lots of filters and conditions. We've changed and created indexes in the database but all of the tests ended up with very expensive queries.
```
menduz, line 75:
```suggestion
Why is `/deployments` endpoint so expensive? It’s an endpoint with lots of filters and conditions. Indexes were created and changed in the database, but all of the tests ended up with very expensive queries.
```

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+
+The snapshots are generated once a period of time (currently 15 min but it could be 6hs) with a full scan of the database which only filters that the overwritten_by field is null, this ensures that all active entities are retried. Then that data is stored in a file, so when requesting the `/snapshots` endpoint you receive the hash of the file with the information and no query to the database is done.
+
+Then `/pointer-changes` retrieves all the deployments done in a period of time, this includes not active entities too. So, the only historical data that may be lost is the data transferred when a Catalyst was down. This endpoint is stored in a separate db and the requests are optimized.
```
menduz, line 79:
```suggestion
Then `/pointer-changes` retrieves all the deployments done in a period of time, this includes not active entities too. So, the only historical data that may be lost is the data transferred when a Catalyst was down. This endpoint is stored in a separate db and the requests are optimized.

A future addition (and enhancement) to the synchronization contemplates the addition of an endpoint similar to `/pointer-changes` that only includes the freshest pointers, ignoring the history in between. For a lighter use of the table and indexes holding the active entities.
```

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+
+
+## Status
+
+Accepted
+
+## Consequences
+
+First of all, to synchronize a new Catalyst from scratch it now takes 6hours while we weren’t able to bootstrap a new Catalyst with the old sync.
```
menduz, line 88:
```suggestion
First of all, to synchronize a new Catalyst from scratch it now takes 6hours while it wasn't possible to bootstrap a new Catalyst with the old sync. Community members gave up after a month of "Bootstrapping" state.
```

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+
+Then, we found less CPU usage on the servers and that implied in increasing the max concurrent amount of users.
```
menduz, line 90:
```suggestion
Then, CPU usage was reduced on the servers and that implied in increasing the max concurrent amount of users and profile deployments.
```

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+
+We are already working on:
+- Removing pg-promise library
+- Store all files in IPFS
```
menduz, line 94: This may be removed from the ADR

docs/ADR-52-content-new-sync.md
```diff
@@ -0,0 +1,107 @@
...
+
+And we identified that the size of the snapshots may increase a lot as more active entities will be active. Another idea that we’ve been discussing is that not all catalysts sync all the entity types.
```
menduz, line 96:
```suggestion
A freshly identified issue with the new design is that the size of the snapshots' size may increase a lot as more active entities will be active. A possible mitigation or optimization could be that not all catalysts sync all the entity types, making specialized catalysts for either scenes+wearables and others for profiles.
```

pentreathm (approved):

menduz (approved):

PR #41, opened 2022-01-27
Feat: adr 58 third party integration v2

nachomazzara:

PR #42, opened 2022-01-28
feat: ADR-59 Store entity

fzavalia:
ADR for the Store Entity.

Contains the schema and an explanation of all properties.

Also a way on how to fetch a store from the catalyst and an example of the payload.


fzavalia:
Skipped from 57 to 59 because at the moment it seems that 58 is being worked here https://github.com/decentraland/adr/pull/41

guidota (approved):
LGTM 👍

nachomazzara (approved):

PR #48, opened 2022-03-10
feat: Add ADR 62 - Merkle proofed entities

LautaroPetaccio:
This PR adds a new ADR to make any entities verifiable using Merkle Trees.


docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,237 @@
...
+      "0x98c363c32c7b1d7914332efaa19ad2bee7e110d79d7690650dbe7ce8ba1002a2",
+      "0x0bd810301fbafeb4848f7b60a378c9017a452286836d19a108812682edf8a12a",
+      "0x1533c6b3879f90b92fc97ec9a1db86f201623481b1e0dc0eefa387584c5d93da",
+      "0x31c2c3dbf88646a964edd88edb864b536182619a02905eaac2a00b0c5a6ae207",
+      "0xc2088dbbecba4f7dd06c689b7c1a1e6a822d20d4665b2f9353715fc3a5f0d588",
+      "0x9e191109e34d166ac72033dce274a82c488721a274087ae97b62c9a51944e86f",
+      "0x5ff2905107fe4cce21c93504414d9548f311cd27efe5696c0e03acc059d2e445",
+      "0x6c764a5d8ded16bf0b04028b5754afbd216b111fa0c9b10f2126ac2e9002e2fa"
+    ],
+    "hashingKeys": [
```
LautaroPetaccio, line 158: Should we specify a mechanism to pick specific keys of a nested object?
For example, the property data contains the property `replaces`, but I would like to not hash it at all. Should we be able to describe the whole list of properties using a syntax like `data.representations`, `data.category`, etc?

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,237 @@
...
+The following is an example on how to generate a Merkle proof for a given entity.
+
+```typescript
+function ipfsHash(metadata, keys): string {
+  return hash(JSON.stringify(pick(metadata, keys)));
+}
+
+// Using the entity, the keys to be hashed and the other node hashes, build the merkle proof for the entity and return a new proofed entity.
+function buildEntityMetadataWithMerkleProof(baseEntity, keys, otherNodeHashes) {
+  const entityHash = ipfsHash(entity, keys);
```
LautaroPetaccio, line 188: The name `ipfsHash` is a placeholder for the new hashing algorithm.

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,237 @@
...
+      "image",
+      "thumbnail",
+      "data",
+      "i18n",
+      "createdAt",
+      "updatedAt",
+      "metrics"
+    ],
+    "treeHashingMethod": "sha256",
+    "entityHashingMethod": "IPFS",
```
LautaroPetaccio, line 171: The entity hashing algorithm is different from the tree hashing algorithm, should we have two properties for them?

menduz: Is there a need for different hashing algos? I don't see the point in using IPFS hashing, since it is indeed very complex in comparision with sha256/keccak.

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Proposed solution
+
+In order to avoid changing the entity structure, the proposed solution will leverage the entities' metadata by including a new property, `merkleProof`, that will contain everything
+needed to verify the entity against a Merkle Tree.
+
+The `merkleProof` property will be an object with the following schema:
+
+```json
+{
```
menduz, line 21: could this be written as typescrypt type instead? for legibility

docs/ADR-61-merkle-proofed-entities.md
menduz: I think we are covered by only specifying one key right now to not overly complex the solution

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,237 @@
...
+  }
+}
+```
+
+### Proof generation
+
+The following is an example on how to generate a Merkle proof for a given entity.
+
+```typescript
+function ipfsHash(metadata, keys): string {
```
menduz, line 182: why is this function called ipfsHash? may `hashPartialObject` be a better name?

docs/ADR-61-merkle-proofed-entities.md
menduz, line 188:
```suggestion
  const entityHash = ipfsHash(baseEntity, keys);
```

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,237 @@
...
+function verifyHash(metadata, requiredKeys) {
+  // The keys used to create the hash MUST be present if they're required.
+  assert(
+    requiredKeys.every((key) => metadata.merkleRoot.hashingKeys.includes(key))
+  );
+  const generatedCrcHash = ipfsHash(metadata, metadata.merkleRoot.hashingKeys);
+  // The hash provided in the merkleProof for the entity MUST match the hash generated by the validator.
+  assert(metadata.merkleRoot.entityHash === generatedCrcHash);
+
+  const merkleRoot = getMerkleRoot(metadata.id);
```
menduz, line 218:
```suggestion
  // based on the metadata.id, the merkle root is obtained from a trusted source i.e: the graph, or a smart contract directly
  const merkleRoot = getMerkleRoot(metadata.id);
```

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: I wanted to be explicit on the algorithm to hash the entity as we're describing it in the `merkleProof` property. `ipfsHash` was taken from the name of a function that was used in a repository, I don't remember if it was the Catalyst, to name the new hashing algorithm.

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: Fixed it!

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: I thought it was clearer in JSON, changed it to TS.

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: That's perfect. I've just added a message making this clearer:
```
The property works by whitelisting the keys at root level that will be included in the object to be hashed
```

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: AFIK there are two algorithms to verify a content hash in the Catalyst ([here](https://github.com/decentraland/content-validator/blob/b562d2f68a44041c36b09aa23e909e12c093fdd9/src/validations/access-checker/wearables.ts#L109)) and one of them is called IPFSHash. Since there are two available, I would think that in the future, others would be available too.

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,238 @@
...
+
+Example of a Third Party wearable entity metadata with a merkleProof:
+
+```typescript
+{
+  id: "urn:decentraland:off-chain:base-avatars:aviatorstyle",
+  name: "Aviator Style",
+  description: "aDescription",
+  image: "image.png",
+  thumbnail: "thumbnail.png",
```
LautaroPetaccio, line 111: I haven't included the redundant content information, that is:
```typescript
content: {
  'thumbnail.png': 'someHash'
}
```
Should we include it here or should we assume that it will always be hashed?

docs/ADR-61-merkle-proofed-entities.md
menduz: for the sake of simplicity, and considering we are doing something from scratch, I'd use sha256 plainly. it is broadly available and several orders of magnitude simpler than IPFS CIDs.

does anybody sees this as a blocker?

docs/ADR-61-merkle-proofed-entities.md
menduz: I think that is implementation dependent, "content" will simply be a key after all. Its content can vary from entity to entity. I think the proposal already covers the basics to ensure a good security level.

But yes, for wearables I'd definitely add the `"content": { key: hash,... }`

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,238 @@
+# Merkle proofed entities
+
+## Context and Problem Statement
+
+The Third Party Wearables initiative had the requirement of being able to publish collections containing a big amount of wearables.
+These big amount wearables would require multiple transactions to be sent and approved, incurring in a high cost in gas and in difficult UX experience.
```
nachomazzara, line 6:
```suggestion
This big amount of wearables would require multiple transactions to be sent and approved, incurring a high cost in gas and in difficult UX experience.
```

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,238 @@
...
+      "type": "string"
+    },
+    "entityHash": {
+      "type": "string"
+    }
+  }
+}
+```
+
+Where `proof` is the array of hashes of nodes needed for the verification, `index` is the index of the entity in the Merkle Tree, `hashingKeys` is the array of keys of the entity metadata that
```
nachomazzara, line 52:
```suggestion
Where `proof` is the array of nodes needed for the verification, `index` is the index of the entity in the Merkle Tree, `hashingKeys` is the array of keys of the entity metadata that
```

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,238 @@
...
+      "description",
+      "image",
+      "thumbnail",
+      "data",
+      "i18n",
+      "createdAt",
+      "updatedAt",
+      "metrics",
+    ],
+    treeHashingMethod: "sha256",
```
nachomazzara, line 171: I'm not 100% sure that the method is sha256.

Screen Shot 2022-03-10 at 19 46 10

> Ethereum uses Keccak-256 in a consensus engine called Ethash. Keccak is a family of hash functions that eventually got standardized to SHA-3 (SHA256 is part of a family of hash functions called SHA-2). Ethereum called it Keccak instead of SHA-3 as it has slightly different parameters than the current SHA-3. Colloquially, Ethereum mining is never called Keccak mining because Ethash utilizes mixhashes in a DAG, which is different from Hashcash proof-of-work. Some blockchains utilize Keccak in the same way Bitcoin uses SHA-256, but these are not as common.


from https://www.quora.com/Whats-the-difference-between-SHA-256-and-Keccak-256-Which-one-should-be-used-on-Ethereum-for-bitcoin-like-mining

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,238 @@
...
+### Proof generation
+
+The following is an example on how to generate a Merkle proof for a given entity.
+
+```typescript
+function ipfsHash(metadata, keys): string {
+  return hash(JSON.stringify(pick(metadata, keys)));
+}
+
+// Using the entity, the keys to be hashed and the other node hashes, build the merkle proof for the entity and return a new proofed entity.
```
nachomazzara, line 187: Do you want to add here the usage of the lib to generate the tree? I think that it is missing how from the first node `ipfsHash(metadata, keys)` you build the three and then build the entity. 

[lib](https://www.npmjs.com/package/@dcl/content-hash-tree)

docs/ADR-61-merkle-proofed-entities.md
```diff
@@ -0,0 +1,238 @@
...
+    requiredKeys.every((key) => metadata.merkleRoot.hashingKeys.includes(key))
+  );
+  const generatedCrcHash = ipfsHash(metadata, metadata.merkleRoot.hashingKeys);
+  // The hash provided in the merkleProof for the entity MUST match the hash generated by the validator.
+  assert(metadata.merkleRoot.entityHash === generatedCrcHash);
+  // Based on the metadata.id, the merkle root is obtained from a trusted source i.e: the graph, or a smart contract directly
+  const merkleRoot = getMerkleRoot(metadata.id);
+  // Verify if the entity belongs to the Merkle Tree.
+  asset(
+    isProofValid(
```
nachomazzara, line 222: Just in case you want to add it. This should use the `verifyProof` from the [content hash tree lib](https://www.npmjs.com/package/@dcl/content-hash-tree)

nachomazzara (changes requested):

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: I understand, but the IPFS algorithm is what is used in the catalyst to do the entity validations, the team also said that algorithm is the one we should use from now on, should the validation procedure add the sha256 one?

docs/ADR-61-merkle-proofed-entities.md
guidota: Should we include the `content` object to this hashing method like we do for ADR32?

menduz: I think we could, but then the entire ADR would work over the entity and not over the metadata, I see value in two approaches with their own implications:
1. Include the `entity.content` among the `entity.metadata*keys`
2. Manually allow a custom key mimicking the `entity.content: Record<Filename, CID>` inside the `metadata.content` and perform the validation manually in every case.

The first one inherits the security for every deployed entity, the second one 1) allows for more flexibility and 2) allows this ADR to be used (to validate) things different than deployed entities. It simply works on records/objects.

Whatever you decide I'm ok with it

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: Added the code to generate the merkle tree.

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: Changed it to use the same function found in the lib.

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: It's not sha256 as you said.
The properties that specify the algorithms were removed.

docs/ADR-61-merkle-proofed-entities.md
LautaroPetaccio: To be a bit more sure about how we're going to handle this:
- The UI builds the entity by adding the contents property to the metadata, the UI hashes the metadata and gets the content hash that will later be used to build the merkle root.
- The Catalysts take the entity contents and the entity metadata and build a new metadata object recomputing the contents object (it discards it and recomputes it again with the contents in the entity) to later hash it and verify the proof with the merkle root.
Am I right?

LautaroPetaccio: The other thing we could do is to keep the ADR32 format, that is the object with `contents` and `metadata` and if the entity doesn't have any contents, we can leave the object empty.

LautaroPetaccio: We've talked with @nachomazzara and @menduz about the implementation an how the Catalyst is going to validate the Third Party entities:
1. The UI builds the entity by adding the contents property to the metadata, the UI hashes the metadata and gets the content hash that will later be used to build the merkle root.
2. The Catalysts take the entity contents and the entity metadata and check that the contents and the metadata.contents are the same set of contents, that is, they have the same amount of files, with the same name and the same hashes. After doing this operation, the Catalysts will hash the entity to compare it with the sent hash (for a simple check) to later check it against the proof.

docs/ADR-62-merkle-proofed-entities.md
```diff
@@ -0,0 +1,238 @@
...
+      "0x0bd810301fbafeb4848f7b60a378c9017a452286836d19a108812682edf8a12a",
+      "0x1533c6b3879f90b92fc97ec9a1db86f201623481b1e0dc0eefa387584c5d93da",
+      "0x31c2c3dbf88646a964edd88edb864b536182619a02905eaac2a00b0c5a6ae207",
+      "0xc2088dbbecba4f7dd06c689b7c1a1e6a822d20d4665b2f9353715fc3a5f0d588",
+      "0x9e191109e34d166ac72033dce274a82c488721a274087ae97b62c9a51944e86f",
+      "0x5ff2905107fe4cce21c93504414d9548f311cd27efe5696c0e03acc059d2e445",
+      "0x6c764a5d8ded16bf0b04028b5754afbd216b111fa0c9b10f2126ac2e9002e2fa",
+    ],
+    hashingKeys: [
+      "id",
```
nachomazzara, line 153: `content` is missing here?

LautaroPetaccio: Yes, adding it!

nachomazzara (approved):

guidota (approved):

PR #54, opened 2022-04-04
[Content Server] URN prefix search

agusaldasoro:


docs/ADR-65-urn-prefix-search.md
```diff
@@ -0,0 +1,104 @@
...
+
+Accepted
+
+## Consequences
+
+### Don't accept wearables deployments with more than one pointer (URN)
+
+Catalyst Issue: [@catalyst#884](https://github.com/decentraland/catalyst/issues/884)
+
+To support the sync of Base and L1 wearables the check of resolving to the same URN was added, that's okay for those, but it's not accurate for L2 (Polygon) wearables. So, for new wearable deployments exactly one pointer is allowed.
```
menduz, line 40: I don't consider this is a blocker or a need. In the worst case scenario, entities with many pointers will appear in two different queries for different prefixes. The generalization of the problem is to look for entities with prefix `-` that will render a response of around 45k LAND starting with a negative number. Many of those will be repeated, because scenes share many lands

menduz (changes requested):

docs/ADR-65-urn-prefix-search.md
agusaldasoro: For the Platform team, it's not a blocker, as long as the client implements the logic to query an item twice, as it's not the Content Server's responsibility to implement the logic to retrieve both entities when matched.

What do you think @nachomazzara?

menduz: I think the query must dedup the entities pointed twice. In SQL that should be inverting an inner join order and adding a distinct.

```sql
-- from this
SELECT 
  d.* 
FROM active_pointer a
INNER JOIN deployments d  ON a.pointer IN d.pointers -- pseudo SQL
WHERE ...
```

```sql
-- to this
SELECT DISTINCT
  d.* 
FROM deployments d 
INNER JOIN active_pointer a ON a.pointer IN d.pointers -- pseudo SQL
WHERE ...
```

agusaldasoro: If the deployment in the deployments table contains two pointers, then in the active_pointer there will be two rows. Is that enough?

menduz: Yes, but the endpoint should return only one entry with two pointers in that case. Are we in the same page?

agusaldasoro: The endpoint response format is:

```
[
{
  "pointer": "onePointer",
  "entityId": "hash"
}
]
```

So, it won't retrieve both pointers. It's client's side responsability

pentreathm (approved):

PR #57, opened 2022-04-06
feat: ADR 68 Configuring wearables & ADR 69 Importing wearables

LautaroPetaccio:
This PR describes two new ADRs:
1. The first ADR describes a new `wearable.json` file, that has the responsibility to describe a wearable. The idea of the ADR is to define a format that can be used anywhere to describe the properties that a wearable has.
2. The second ADR describes a new `builder.json` file, that has the responsibility to describe the information of a wearable (or an item) in the Builder platform.

These two ADRs split the responsibility of the current `asset.json` file and define more properties to fully describe both resources.


docs/ADR-65-importing-wearables.md
```diff
@@ -0,0 +1,200 @@
...
+4. The collection where the wearable is created into can't be specified.
+
+These downsides make the current `asset.json` configuration file not suitable for importing wearables.
+
+## Proposed solution
+
+To solve the downsides of the current format, a new format for the `asset.json`, with the following schema is proposed:
+
+```json
+{
```
menduz, line 53: It is _really_ hard to understand json schemas, can this be changed to typescript types?

AjimenezDCL:
Just a quick heads-up @LautaroPetaccio , I just merged a 65 and 66 ADRs. Can you move yours to 67?

cazala:
This looks good to me, my only concern is the name of the file. We already have an `asset.json` for Builder assets, which is pretty much the same concept, but a different schema. Once we support Emotes, we could also want some sort of configuration via json like this, and the schema will be different too. Should we name the file something more specific? like `asset.json` for builder assets and smart assets, `wearable.json` for wearables and smart wearables, `emote.json` for emotes and so forth?

docs/ADR-67-importing-wearables.md
```diff
@@ -0,0 +1,200 @@
...
+1. Adding the `representations` property, where each representation can be detailed by specifying the `body shape` (`male`, `female` or `both` which will be used to build two representations), the `main file` (the file path of the main 3D asset that is used in the representation) and the `contents` (a list of file paths belonging to the content) of each representation.
+2. Adding the `tags` property to be able to define tags for the wearable.
+3. Defining the `hides`, `replace` properties for the whole wearable and along with the representations' data the `overrideHides` with `overrideReplaces` properties, solving the issue of customizing how the wearable and its representations hide or replace other wearables.
+4. Adding the `collectionId` property to specify in which collection the wearable should be imported to.
+
+The following example shows how a wearable could be described using the new `asset.json` format:
+
+```json
+{
+  "id": "f12313b4-a76b-4c9e-a2a3-ab460f59bd67",
```
gonpombo8, line 176: `assetType` is missing right ?

docs/ADR-65-importing-wearables.md
LautaroPetaccio: Changed it!

LautaroPetaccio:
We've taken a look at the model with the @gonpombo8 and @leanmendoza and we've made some changes to it.
We're removing the `assetType` and `menuBarIcon` properties, as the first one is not going to be needed anymore because we're renaming the file to `wearable.json` as @cazala recommended and the other property will be moved into another file.

docs/ADR-65-importing-wearables.md
menduz: it still looks like a json schema?

menduz:
Help me understand, after this change, where will the explorer takes the `menuBarIcon` from?

LautaroPetaccio:
We can detect if we're uploading a smart wearable so at the time of deploying the item to the Catalyst, we can read that `scene.json` and deploy the smart item with that property.
Just to be clear, we're not deploying any wearables at the moment with the `menuBarIcon` property.

gonpombo8 (approved):

docs/ADR-67-importing-wearables.md
```diff
@@ -0,0 +1,206 @@
...
+      ],
+    },
+    "wearable-representation": {
+      type: "object",
+      properties: {
+        bodyShapes: {
+          type: "array",
+          items: {
+            type: "string",
+            enum: ["both", "female", "male"],
```
cazala, line 126: uh this is already an array, so having `both` in the enum doesn't make much sense, instead it would be an array with `male` and `female`. Also, instead of using plain `female` and `male`, maybe we should use the actual urn of the body shape ([link](https://github.com/decentraland/common-schemas/blob/main/src/platform/wearables/wearable-body-shape.ts#L8-L11)), which is what is used by the `WearableRepresentation` type ([link](https://github.com/decentraland/common-schemas/blob/main/src/platform/wearables/representation.ts#L11)), that looks like: `urn:decentraland:off-chain:base-avatars:BaseMale`

We already have 3 **different** schemas for representations ([1](https://github.com/decentraland/common-schemas/blob/main/src/platform/wearables/representation.ts#L13), [2](https://github.com/decentraland/common-schemas/blob/main/src/dapps/preview/representation-definition.ts#L7), [3](https://github.com/decentraland/kernel/blob/62dcd0f5f971198490a2630332342bd78b70dbb4/packages/shared/catalogs/types.ts#L75)) lets not create a 4th one 😅

cazala (changes requested):

docs/ADR-67-importing-wearables.md
LautaroPetaccio: You're right! the type is wrong, it should be a string with any of those values ('both', 'female' or 'male'). The rationale is to have something less verbose for the users to write when writing this file to import their wearables.
I'll change it to be a string.

cazala (approved):

docs/ADR-65-importing-wearables.md
LautaroPetaccio: I'm sorry, I thought that you mean converting the markdown from JSON to TS. My bad. I've just converted it into a TS type.

docs/ADR-67-configuring-wearables.md
```diff
@@ -0,0 +1,169 @@
...
+  replaces: WearableCategory[];
+  /** Tags to identify the wearable */
+  tags: string[];
+  /** Representations of the wearable */
+  representations: WearableRepresentation[];
+}
+
+type WearableRepresentation = {
+  /** Body shape of the representation */
+  bodyShape: BodyShape;
```
cazala, line 81: This type is very similar to the `WearableRepresentation` that exists in the commons schemas repo: https://github.com/decentraland/common-schemas/blob/main/src/platform/wearables/representation.ts#L10-L16

Expect that the one in common schemas uses an array of urns instead of the enum `male`, `female`, `both`. I think it would be better to stick to the type we are already using even tho dealing with urns is more cumbersome for the sake of consistency. Also this design would be more future-proof if at some point more body shapes are supported.

cazala (changes requested):

docs/ADR-67-configuring-wearables.md
LautaroPetaccio: I've changed the whole schema to be identical to the wearable entitiy metadata.

cazala (approved):

PR #61, opened 2022-05-16
docs: add revert and break to git style guide

manumena:


menduz:
Won't it generate confusions with semantic release projects? This ADR is organization-wide @manumena

PR #66, opened 2022-06-07
feat: adr 74 - add emote schema

pedrotambo:


docs/ADR-74-add-emote-schema.md
```diff
@@ -0,0 +1,142 @@
+# Add Emote schema and versioning
+
+## Context and Problem Statement
+
+Currently, `Emotes` are being deployed as `Wearables` in the Catalysts. Even though they are very similar, they are not. For instance, there are emotes deployed with the arbitrary category "hat" just to pass the validations.
+Secondly, we are introducing some kind of versioning within the emotes metadata. This way, we can modify the metadata as we see fit but also keep track of these changes so we can validate the emotes over time. For instance, if we start a new Catalyst from scratch, it will receive entities from differenet timestamps and we have to validate them with the validations that **_where present at that timestamp_**.
```
marianogoldman, line 6:
```suggestion
Secondly, we are introducing some kind of versioning within the emotes metadata. This way, we can modify the metadata as we see fit but also keep track of these changes so we can validate the emotes over time. For instance, if we start a new Catalyst from scratch, it will receive entities from different timestamps and we have to validate them with the validations that **_where present at that timestamp_**.
```

docs/ADR-74-add-emote-schema.md
```diff
@@ -0,0 +1,142 @@
...
+    category: WearableCategory;
+  };
+};
+```
+
+### Changes involved
+
+##### 1. Share common properties
+
+Share some common properties between Wearables/Emotes that we expect not to change in the feature, so there will be no versioning for them (at least in the way we will see later). Move content?: Record<string, string> to ThirdPartyProps as they're only used in that case. Compared to the current schema, all properties are present except the data property (that will go in the versioned part) and content (that is removed):
```
marianogoldman, line 41:
```suggestion
Share some common properties between Wearables/Emotes that we expect not to change in the feature, so there will be no versioning for them (at least in the way we will see later). Move `content?: Record<string, string>` to `ThirdPartyProps` as they're only used in that case. Compared to the current schema, all properties are present except the `data` property (that will go in the versioned part) and `content` (that is removed):
```

marianogoldman (approved):

docs/ADR-74-add-emote-schema.md
```diff
@@ -0,0 +1,142 @@
...
+```typescript
+function validate(emote) {
+   adrXX = identifyADR(emote.timestamp)
+   if (emote does not have the field 'emoteData' + ardXX) return false
+   Emote.schema.validate(emote)
+}
+```
+
+\*under the assumption that the common properties do not change.
+This would give us the possibility to iterate schema changes quicker and also easily identify the versions over the time along with its reasons explained in the specific ADR.
```
agusaldasoro, line 140:
```suggestion
This would give the possibility to iterate schema changes quicker and also easily identify the versions over the time along with its reasons explained in the specific ADR.
```

docs/ADR-74-add-emote-schema.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Date: TBD
```
agusaldasoro, line 142: in a month? two?

PR #86, opened 2022-09-21
docs: add crashbot ADR

manumena:


menduz:
Great. Couple questions:

- The usual PUT semantics seem to be destructive in regards to the previous state, does it work that way or it appends an update every time?
- How are the endpoints going to be secured?

manumena:
@menduz 
1. It would override the previous value for every non-empty field. I'm looking for a way to show a modal in slack with the current incident's data to give a more "editing" experience.
2. I was thinking using signed secrets https://api.slack.com/authentication/verifying-requests-from-slack which is the recommended way for verifying requests, but now I'm testing enabling socket mode (it only requires providing the app tokens when initializing the app). Socket mode would mean we can get rid of `/create`, `/update`, and `/close` endpoints, maintaining only the `/list`.

menduz:
1. Ok, thanks for the clarification all incident managers are append-only, please let's continue that way. Every time a update is added it doesn't delete the previous one.
2. Roger!, would we have internal endpoints to use with curl for debugging or manual editing?

manumena:
1. Understood!
2. In terms of debugging I think we are covered, we can run it and debug it locally in production or test workspace, we will have logs and metrics provided by wkc. But for manual editing we need some endpoints (unless we resolve it by querying the db). I would leave those for a second stage of the project

menduz:
2. I'd be super explicit with the "contingency plans" in the ADR for editing the DB directly.. if the status page is public and we introduce typos, we should have a way to fix them

manumena:
Let's talk about this in our next meeting

ADR/ADR-82-crashbot.md
```diff
@@ -55,61 +55,40 @@ stateDiagram-v2
...
+        contact: string,
+        rca_link: string
       }
     ],
     closed: [
       {
+        id: number,
+        update_number: number,
+        modified_at: string,
+        modified_by: string
```
menduz, line 83:
```suggestion
```

Should we trait the person behind the report as anonymous?

manumena: Every other incident manager keep track of who is making the changes. It is useful whenever you need to ask about the incident or its changes

menduz: But that information is protected, not publicly available

manumena: Great point. We could store it for future usage but not expose it then, what do you think?

PR #90, opened 2022-09-26
feat: RFC-9 On-chain validators

nachomazzara:
RFC for removing the usage of Thegraph on behalf of smart contracts in the content-server


RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,261 @@
...
+
+#### Wearables & Emotes
+
+Each entity has a `wearables` and `emotes` in the property `entity->metadata->avatars`. Wearables are being stored with an array format like `[wearable_urn_1, wearable_item_urn_2, ..., wearable_urn_N]` and Emotes as an array as well but with objects `[{slot: 1, urn: wave}, {slot: 2, urn: emote_item_urn_1}]`.
+Profiles are stored with the item's urn instead of the NFT itself urn. E.g: `urn:decentraland:matic:collections-v2:0x875146d1d26e91c80f25f5966a84b098d3db1fc8:1` (item); `urn:decentraland:matic:collections-v2:0x875146d1d26e91c80f25f5966a84b098d3db1fc8:1:1` (token id/NFT). Checking if a user has a specific item doesn't scale. The ERC721 standard has a method `ownerOf(tokenId: uint256)` to check who is the owner of the NFT. All the profiles are using only item ids instead of the specific NFT. _Remember that NFTs are items kind for Decentraland Collections_. To perform on-chain checks we should perform some changes to how we save profiles and how to validate old profile deployments.
+
+- Old Profile Deployments: Use the Decentraland Address to signal profile deployments in order to consider them valid without the need to perform further validations.
```
agusaldasoro, line 58: Does this imply re-deploying old profiles with a Decentraland Address? Is there a way to continue validating old profiles the way that it's done today?

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,261 @@
...
+  - `decentraland:{protocol}:collections-v2:{contract(0x[a-fA-F0-9]+)}:{itemId}`
+  - `decentraland:{protocol}:collections-v1:{contract(0x[a-fA-F0-9]+)}:{itemId}:{tokenId}` -> new
+  - `decentraland:{protocol}:collections-v1:{collection-name}:{itemName}:{tokenId}` -> new
+  - `decentraland:{protocol}:collections-v2:{contract(0x[a-fA-F0-9]+)}:{itemId}:{tokenId}` -> new
+
+  The **Explorer** should show the items grouped by item kinds but allow the user to select the specific NFT to add to their profile. Once the users save their profile for the first time once the ADR is running effectively, a process in the client should select the first nft for each item kind saved in the profile. E.g: if the user has this item `urn:decentraland:ethereum:collections-v1:wonderzone_steampunk:steampunk_jacket` selected in their profile but he has 10 of them with the token ids from `1` to `10`, the explorer should select the first (`urn:decentraland:ethereum:collections-v1:wonderzone_steampunk:steampunk_jacket:1`) one and replace it. For off-chain wearables, we should not do any kind of validation.
+
+  The **Kernel** and the **Catalysts** should accept and return all the NFTs.
```
agusaldasoro, line 71: What does this mean? And on which cases/endpoints?

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,261 @@
...
+
+#### Wearables & Emotes
+
+Each entity has a `wearables` and `emotes` in the property `entity->metadata->avatars`. Wearables are being stored with an array format like `[wearable_urn_1, wearable_item_urn_2, ..., wearable_urn_N]` and Emotes as an array as well but with objects `[{slot: 1, urn: wave}, {slot: 2, urn: emote_item_urn_1}]`.
+Profiles are stored with the item's urn instead of the NFT itself urn. E.g: `urn:decentraland:matic:collections-v2:0x875146d1d26e91c80f25f5966a84b098d3db1fc8:1` (item); `urn:decentraland:matic:collections-v2:0x875146d1d26e91c80f25f5966a84b098d3db1fc8:1:1` (token id/NFT). Checking if a user has a specific item doesn't scale. The ERC721 standard has a method `ownerOf(tokenId: uint256)` to check who is the owner of the NFT. All the profiles are using only item ids instead of the specific NFT. _Remember that NFTs are items kind for Decentraland Collections_. To perform on-chain checks we should perform some changes to how we save profiles and how to validate old profile deployments.
```
pedrotambo, line 56: I don't understand why it's needed the `tokenId` without TheGraph, but with TheGraph the `itemId` it's enough. What does the TheGraph different?

cazala:
Would this change require the use of an archive node for the RPC? Since for syncing deployments we would need to read the state of the chain at a certain block in the past?

RFC/RFC-9-on-chain-validators.md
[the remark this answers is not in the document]
pedrotambo: Profile `emotes` and `wearables` are stored ... ?

RFC/RFC-9-on-chain-validators.md
pedrotambo, line 58: Do you mean to "don't check items ownership for `profiles` deployed with DCL address and its timestamp < `ADR_TIMESTAMP_RISED_FROM_THIS_RFC`"?

RFC/RFC-9-on-chain-validators.md
cazala: The NFT entity in the subgraph has an `itemId` field. So I guess we query something like:

```
nfts(where: { owner: "0xuser", itemId: "0" }) { ... }
```

And if that query returns _any_ nft we know you at least own an NFT of that item (it doesn't matter which one particularly since they are all the same).

When querying the contract you can only query ownership of a token `ERC721.ownerOf(tokenId)`, there's no method that translates 1:1 to that query (asking if the users owns any nft of a given item).

pedrotambo:
Currently there's two endpoints `/wearables-by-owner` and `/emotes-by-owner` in Lambdas that asks for all the items owned by an owner address. These endpoints make usage of the TheGraph query:
```
  nfts(where: {owner: $owner, id_gt: $start}, first: $first) {
    id
    urn
    collection {
      isApproved
    }
  }
```
How would/could that work with this change?

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,261 @@
...
+
+Remove the usage of subgraphs on behalf of smart contracts to perform validations when deploying entities (access-checkers). This removes the dependency with subgraphs and creates an immutable and decentralized way of validating entity deployments that need to query blockchain information. The content server will still need to use an RPC provider but this RPC is also used nowadays to validate smart contract wallets' deployments.
+
+The final goal is to have a set of different upgradable smart contracts managed by the DAO needed that emulates the subgraphs.
+
+This document also describes a migration approach needed to make profile validations possible.
+
+## Proposed solution
+
+Subgraphs are being used by the content server to get:
```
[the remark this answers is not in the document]
nachomazzara: + 1. Will add it

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,261 @@
...
+
+- Scenes
+- Collections items
+- ThirdParty Wearables
+- Ethereum & Polygon block numbers
+- Profiles
```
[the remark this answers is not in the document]
nachomazzara: But they are happening when the profile is deployed, right?

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,261 @@
...
+
+The following sections will describe how to use only RPC calls and smart contracts to validate deployments by entity type.
+
+### Ethereum & Polygon block numbers
+
+Each entity deployment has a timestamp. We get the block number for that timestamp by using [subgraphs](https://github.com/decentraland/ethereum-blocks) per each chain. It is not possible to get a block number by timestamp directly with only one RPC request because timestamps are just part of the block metadata. In order to remove this dependency, we need extra work:
+
+- A) Use a [binary search](https://github.com/nachomazzara/eth-block-timestamp) or any other new way to get the closest block for a given timestamp.
```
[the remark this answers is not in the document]
nachomazzara: @marianogoldman is working on a [custom implementation](https://github.com/decentraland/block-indexer) for this. Not sure how will work

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,261 @@
...
+- B) Start adding block numbers to the deployments. For old deployments, we can create an immutable file with all the blocks per each timestamp used for previous deployments. This file will have a fixed size because new deployments will have the block number directly inside the deployment.
```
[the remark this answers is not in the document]
nachomazzara: Yes! Every client can get the latest block number. Also for validate the deployment we have the 5 minutes window

RFC/RFC-9-on-chain-validators.md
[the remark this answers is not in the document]
nachomazzara: @pedrotambo and @agusaldasoro. In the profiles, we are storing the item's urn not the NFT urn. The suggestion is not valid. How can I help to remove the confusion?

RFC/RFC-9-on-chain-validators.md
nachomazzara: I would add something that I forgot which is a disclaimer about this section. We will need another RFC or ADR for this part. IMO we should find a solution where we can avoid re-deploying all the profiles again. We will duplicate the storage used if not.

nachomazzara: Exactly!

RFC/RFC-9-on-chain-validators.md
nachomazzara: The lambdas will return the NFTs and not the items anymore. The kernel needs to know that and also @POST new profiles with NFTs and not items.

nachomazzara:
> I understand that using the specific NFT to identify collections in profiles is necessary, anyway the changes needed to support that in The Content Server and in Lambdas Server should be considered as part of the effort of this proposal.

100% We need to write another RFC or ADR for that. I will add a disclaimer to this document about that part. Basically is the part that I would like to receive more feedback about how to implement it and double check if I am missing something

RFC/RFC-9-on-chain-validators.md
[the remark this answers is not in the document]
cazala: Each `Profile` entity stores the `wearables` and `emotes` equipped by the user. Wearables are stored under `entity.metadata.avatar.wearables` as an array of URNs, and Emotes are stored under `entity.metadata.avatar.emotes` as an array of `{ slot: string, urn: string }`. In both cases the URNs used are the **item's** urn, not the **token's** urn. Checking if a user owns a specific item does not scale, since it would require looping within a contract method (users that own too many nfts or collections that are too big could cause this to lag or timeout) or making several RPC calls between the catalyst and the ethereum node (which also would not scale well). We should start storing the **token's** urn within the profiles to make the access checks possible in a single RPC call.

nachomazzara:
> Would this change require the use of an archive node for the RPC? Since for syncing deployments we would need to read the state of the chain at a certain block in the past?
> 
> Dunno if our providers already give us archive node access, maybe they do.

Yes! they are mandatory at the moment as far as I know. Checking smart contract signatures to sync deployments between catalysts needs an archive node to check if the signer was part of the contract wallet at the time of that deployment.  @menduz @agusaldasoro  @pentreathm correct me if I am wrong

nachomazzara:
> Currently there's two endpoints `/wearables-by-owner` and `/emotes-by-owner` in Lambdas that asks for all the items owned by an owner address. These endpoints make usage of the TheGraph query:
> 
> ```
>   nfts(where: {owner: $owner, id_gt: $start}, first: $first) {
>     id
>     urn
>     collection {
>       isApproved
>     }
>   }
> ```
> 
> How would/could that work with this change?

Lambdas are not part of the protocol and we can still use Thegraph or any NFT API provider. The change that we will need to perform is to use the NFT urn instead of the item urn. In the subgraphs,  `nft.urn` returns the item urn instead of the NFT one 🤕

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,273 @@
...
+
+Subgraphs are being used by the content server to check the access to:
+
+- LANDs
+- Collections items
+- ThirdParty Wearables
+- Ethereum & Polygon block numbers
+- Decentraland Names
+
+There are two type of deploymenys:
```
menduz, line 33:
```suggestion
There are two type of deployments:
```

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,273 @@
+---
+layout: doc
+rfc: 9
+date: 2022-09-28
+title: On-chain validators
+authors:
+  - imazzara
```
menduz, line 7:
```suggestion
  - nachomazzara
```
Github user

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,273 @@
...
+
+The following sections will describe how to use only RPC calls and smart contracts to validate deployments by entity type.
+
+### Ethereum & Polygon block numbers
+
+Each entity deployment has a timestamp. We get the block number for that timestamp by using [subgraphs](https://github.com/decentraland/ethereum-blocks) per each chain. It is not possible to get a block number by timestamp directly with only one RPC request because timestamps are just part of the block metadata. In order to remove this dependency, we need extra work:
+
+- A) Use a [binary search](https://github.com/nachomazzara/eth-block-timestamp) or any other new way to get the closest block for a given timestamp.
+- B) Start adding block numbers to the deployments. For old deployments, we can create an immutable file with all the blocks per each timestamp used for previous deployments. This file will have a fixed size because new deployments will have the block number directly inside the deployment. The client can get the latest block number. The content server will keep using the 5 minutes window for validate deployments.
```
menduz, line 45: I think this is un-viable

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### Wearables & Emotes
+
+Each Profile entity stores the wearables and emotes equipped by the user. Wearables are stored under `entity.metadata.avatar.wearables` as an array of items' URNs, and Emotes are stored under `entity.metadata.avatar.emotes` as an array of { slot: string, urn: string }. In both cases, the URNs used are the item's urn, not the token's urn. Checking if a user owns a specific item does not scale, since it would require looping within a contract method. Users that own too many NFTs or collections that are too big could cause this to lag or timeout. Another alternative could be making several RPC calls between the catalyst and the Ethereum nodes, which also would not scale well. We should start storing the token's urn within the profiles to make the access checks possible in a single RPC call. The ERC721 standard has a method `ownerOf(tokenId: uint256)` which serves to get the owner of the NFT.
+
+This document will propose two ideas on how the old profiles can be migrated but an new RFC or ADR should be needed describing the final approach.
+
+- ##### Alternative 1
```
menduz, line 64: item list -> header
```suggestion
##### Alternative 1
```

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,273 @@
...
+
+  Use the Decentraland Address to signal profile deployments in order to consider them valid without the need to perform further validations. It can be a new field to the profiles schema to avoid re-deploying the profiles by the Decentraland address which can duplicate the storage. If duplicating the profiles is the only alternative, then old ones can be removed.
+
+- ##### Alternative 2
```
menduz, line 68:
```suggestion
##### Alternative 2
```

menduz:
@guidota could you check how this plays with the ADR of the validations for entity deployments?

RFC/RFC-9-on-chain-validators.md
pedrotambo: Yeah, was wondering what does TheGraph to know that a user owns `some` nft of that `itemId`. I imagine that we could get all the `tokenId`'s for that `itemId` and query `ERC721.ownerOf(tokenId)` for each one, but that doesn't scale right?

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,273 @@
...
+  Build a [Merkle tree with every profile hash](https://github.com/decentraland/content-hash-tree) and submit the root to a smart contract. A static file must be added to the repository with all the proofs needed to check if a deployment is part of the tree. The file with the proofs may weight [~100MB for 1M of profiles](https://github.com/decentraland/content-hash-tree#generate-a-tree-with-content-hashes).
+
+For new profiles, a new ADR with a date to make it effective to start receiving profiles with an extended urn with the token id at the end. This implies changes in the Explorer, Kernel, Catalysts, and URN Resolver. New urn resolvers
+
+- `decentraland:{protocol}:collections-v1:{contract(0x[a-fA-F0-9]+)}:{itemName}`
+- `decentraland:{protocol}:collections-v1:{collection-name}:{itemName}`
+- `decentraland:{protocol}:collections-v2:{contract(0x[a-fA-F0-9]+)}:{itemId}`
+- `decentraland:{protocol}:collections-v1:{contract(0x[a-fA-F0-9]+)}:{itemId}:{tokenId}` -> new
+- `decentraland:{protocol}:collections-v1:{collection-name}:{itemName}:{tokenId}` -> new
+- `decentraland:{protocol}:collections-v2:{contract(0x[a-fA-F0-9]+)}:{itemId}:{tokenId}` -> new
```
pedrotambo, line 79: One question here, is the next right?
- `decentraland......{itemId}:{tokenId}` <- urns for items in profiles
- `decentraland......{itemId}` <- urns for the item saved in the content-server?

nachomazzara: > `decentraland......{itemId}:{tokenId}` <- urns for items in profiles

urns for NFTs in profile

> decentraland......{itemId} <- urns for the item saved in the content-server?
urns for the item, used when a creator deploys an update to the wearable or emote.

RFC/RFC-9-on-chain-validators.md
agusaldasoro: Okay, for that part I think that we should have documented all the necessary changes in the Content and Lambdas APIs, and also a mechanism to release all those breaking changes.

RFC/RFC-9-on-chain-validators.md
pedrotambo: 👍

RFC/RFC-9-on-chain-validators.md
pedrotambo: Oh okay makes sense now, thanks 👍

pedrotambo:
> > Currently there's two endpoints `/wearables-by-owner` and `/emotes-by-owner` in Lambdas that asks for all the items owned by an owner address. These endpoints make usage of the TheGraph query:
> > ```
> >   nfts(where: {owner: $owner, id_gt: $start}, first: $first) {
> >     id
> >     urn
> >     collection {
> >       isApproved
> >     }
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > How would/could that work with this change?
> 
> Lambdas are not part of the protocol and we can still use Thegraph or any NFT API provider. This document is for removing the usage of subragphs in the content-server only as a starting point. The change that we will need to perform is to use the NFT urn instead of the item urn. In the subgraphs, `nft.urn` returns the item urn instead of the NFT one 🤕

Okay great. I think it'd be worth mentioning that in the RFC/ADR as those endpoints are mentioned.

guidota:
Access validations seem to be aligned with those described in [ADR-51](https://rfc.decentraland.org/adr/ADR-51).

In cases like scenes, we use a `SCENE_LOOKBACK_TIME` and perform the validation twice for each parcel. I wonder if there is room for improvement here and if batching would be possible to avoid multiple calls to the provider.

nachomazzara:
> Access validations seem to be aligned with those described in [ADR-51](https://rfc.decentraland.org/adr/ADR-51).
> 
> In cases like scenes, we use a `SCENE_LOOKBACK_TIME` and perform the validation twice for each parcel. I wonder if there is room for improvement here and if batching would be possible to avoid multiple calls to the provider.

Yes, you can use _Multicall_ which means that you can perform multiple contract calls by just doing one RPC call. I'm curious why validate them twice though.

nachomazzara:
> `/wearables-by-owner` and `/emotes-by-owner`



> > > Currently there's two endpoints `/wearables-by-owner` and `/emotes-by-owner` in Lambdas that asks for all the items owned by an owner address. These endpoints make usage of the TheGraph query:
> > > ```
> > >   nfts(where: {owner: $owner, id_gt: $start}, first: $first) {
> > >     id
> > >     urn
> > >     collection {
> > >       isApproved
> > >     }
> > >   }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > How would/could that work with this change?
> > 
> > 
> > Lambdas are not part of the protocol and we can still use Thegraph or any NFT API provider. This document is for removing the usage of subragphs in the content-server only as a starting point. The change that we will need to perform is to use the NFT urn instead of the item urn. In the subgraphs, `nft.urn` returns the item urn instead of the NFT one 🤕
> 
> Okay great. I think it'd be worth mentioning that in the RFC/ADR as those endpoints are mentioned.

Agree, [Added](https://github.com/decentraland/adr/pull/90/commits/e711bb7befe6af49c0189f305297b9c8a16ef9ee)

RFC/RFC-9-on-chain-validators.md
```diff
@@ -0,0 +1,273 @@
...
+- **Pointer**: `decentraland:{protocol}:collections-v1:{contract(0x[a-fA-F0-9]+)}:{itemName}` (Item URN)
+- **Contracts**: Ethereum Decentraland Collections
+- **Check**: Off-chain check. The only signer allowed to deploy Ethereum wearables is the Decentraland Address. Decentraland is not creating more collections in Ethereum. The ones created are considered **legacy** and the addresses are known beforehand. A list of those addresses can be used to check if the collection address is valid.
+
+### Polygon Items: Wearables & Emotes
+
+- **Entity**: Wearable & Emote
+- **Pointer**: `decentraland:{protocol}:collections-v2:{contract(0x[a-fA-F0-9]+)}:{itemId}` (Item URN)
+- **Contracts**: Polygon Decentraland Collections
+- **Check**: Each collection in Polygon is created by a collection factory. The addresses of the factories are known in advance. For the time of this RFC, Decentraland has only two collection factories. The information needed to perfom the validity of the deployment is the contract address and the item id which should be extracted from the urn, and the content hash and the signer of the deployment. The on-chain validations needed are:
```
marianogoldman, line 252:
```suggestion
- **Check**: Each collection in Polygon is created by a collection factory. The addresses of the factories are known in advance. At the time of this RFC, Decentraland has only two collection factories. The information needed to check the validity of the deployment is the contract address and the item id which should be extracted from the urn, and the content hash and the signer of the deployment. The on-chain validations needed are:
```

PR #93, opened 2022-10-04
RFC-10: realm description

hugoArregui:


RFC/RFC-10-realm-description.md
```diff
@@ -0,0 +1,129 @@
...
+    "publicUrl": "/"
+  }
+}
+```
+
+## Health
+
+If the realm is in healthy state, the endpoint must return a http status 200, otherwise it should return 503. This way just by checking the response http status, a client looking for a realm to connect may decide to connect or not to the realm.
+
+The `healthy` field in the root of the structure) will be true only if `comms.healthy && lambdas.healthy && bff.healthy` is true.
```
menduz, line 98: All the referenced services

RFC/RFC-10-realm-description.md
```diff
@@ -0,0 +1,129 @@
...
+
+This document defines the protocol to describe a realm.
+
+## Realm
+
+We call realm to a given set of services needed for the client to work: current we have comms, content and lambdas. We point to a realm by using an URL, it's a valid realm if it provides a `/about` endpoint in the proper format describing the realm and its status.
+
+## /about
+
+This endpoint should return a json like this:
```
menduz, line 23: Would you add a link to the protocol definition of this type?

RFC/RFC-10-realm-description.md
```diff
@@ -0,0 +1,129 @@
...
+status: DRAFT
+---
+
+# Abstract
+
+This document defines the protocol to describe a realm.
+
+## Realm
+
+We call realm to a given set of services needed for the client to work: current we have comms, content and lambdas. We point to a realm by using an URL, it's a valid realm if it provides a `/about` endpoint in the proper format describing the realm and its status.
```
menduz, line 19: Can you rephrase this removing all the "we"?

menduz (approved):

PR #94, opened 2022-10-04
Update RFC-10-hosted-preview-service.md

agusaldasoro:


RFC/RFC-10-hosted-preview-service.md
```diff
@@ -61,6 +61,53 @@ flowchart TD
     E("Catalyst/Realm") -- avatars -->B
 ```
 
+#### BFF
+
+In order to run, the Client needs to connect to a BFF. In this case, the Preview Content Server will act as BFF implementing the endpoint `/about` documented in RFC-10.
```
pentreathm, line 66: this is RFC-10

agusaldasoro: not anymore haha https://github.com/decentraland/adr/blob/main/RFC/RFC-10-realm-description.md

hugoArregui: too slow :smile:

PR #105, opened 2022-10-17
RFC: P2P adapter

hugoArregui:


RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+title: P2P adapter with routing service
+status: DRAFT
+authors:
+  - hugoArregui
+  - agusAldasoro
+---
+
+## Abstract
+
+This document describes a possible implementation for a P2P adapter using a fallback messaging service. For more information about what and adapter is, check [ADR-81](/ADR/ADR-81-minimum-comms-transport.md).
```
marianogoldman, line 14:
```suggestion
This document describes a possible implementation for a P2P adapter using a fallback messaging service. For more information about what an adapter is, check [ADR-81](/ADR/ADR-81-minimum-comms-transport.md).
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+A P2P adapter is needed in order to provide a light mechanism for comms (compared to [RFC-5: comms service using websocket](/RFC/RFC-5-ws-comms.md)). This adapter should guarantee the delivery of a message to any peer in the network (or cluster, [real-time clustering of users in Islands (ADR-35)](/adr/ADR-35)).
+
+## Approach
+
+This approach is simple, it should be easy to debug and it could be refined in the future for performance if needed. 
+
+This implementation requires two extra services:
+
+- Messaging service: in charge of sending messages between specific peers.
+- Routing service: in charge of keeping track of the connections between peer and given them back routing information.
```
marianogoldman, line 27:
```suggestion
- Routing service: in charge of keeping track of the connections between peers and providing them routing information.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+
+As defined in [ADR-81](/ADR/ADR-81-minimum-comms-transport.md), each peer will known the ids of the peers around them using the information provided by the CommunicationsDirector.
```
marianogoldman, line 29:
```suggestion
As defined in [ADR-81](/ADR/ADR-81-minimum-comms-transport.md), each peer will know the ids of the peers around them using the information provided by the `CommunicationsDirector`.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+
+The basic idea of this implementation is for peers to connect randomly to each other forming a mesh, and reporting their connections to the routing service. The routing service will build routing tables for each peer to connect to every other. Each peer will receive from the routing service the path to all other conected peers. When a peer delivers a message, it will use the path provided by the routing table, and if there is no one, it will fallback to the messaging service.
```
marianogoldman, line 31:
```suggestion
The basic idea of this implementation is for peers to connect randomly to a subset of the others forming a mesh, and reporting their connections to the routing service. The routing service will build routing tables for each peer to connect to every other. Each peer will receive from the routing service the path to all other connected peers. When a peer needs to deliver a message, it will use the path provided by the routing table, and if there is not any, it will fallback to the messaging service.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+
+Although this implementation requires a server, it tries to maximize the usage of P2P connections, while also keeping in mind that peers may run in a constraint enviroment (like a browser) in which not always is possible to have a lot of open connections.
```
marianogoldman, line 33:
```suggestion
Although this implementation requires a server, it tries to maximize the usage of P2P connections, while also keeping in mind that peers may run in a constrained environment (like a browser) in which it is not always possible to have a lot of open connections.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+    Peer3->>Peer1: Establish P2P webrtc connection
+
+    Peer1->>Peer4: Initiate P2P webrtc connection
+    Peer4->>Peer1: Establish P2P webrtc connection
+
+    Peer2->>Peer3: Initiate P2P webrtc connection
+    Peer3->>Peer2: Establish P2P webrtc connection
+```
+
+### 2. Each peer report its connections to the routing service
```
marianogoldman, line 60:
```suggestion
### 2. Each peer reports its connections to the routing service
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+    participant RS as Routing Service
+
+    Peer1->>RS: connected to [peer2, peer3, peer4]
+    Peer2->>RS: connected to [peer1, peer3]
+    Peer3->>RS: connected to [peer1, peer2]
+    Peer4->>RS: connected to [peer1]
+    Peer5->>RS: connected to []
+```
+
+### 3. The routing service create a routing table for each peer
```
marianogoldman, line 78:
```suggestion
### 3. The routing service creates a routing table for each peer
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+```mermaid
+sequenceDiagram
+    participant Peer1
+    participant Peer2
+    participant Peer3
+    participant Peer4
+    participant Peer5
+    participant RS as Routing Service
+
+    RS->>Peer1: { peer2: [], peer3: [], peer4: [] }
```
marianogoldman, line 93: So, RS never tells anyone about routes (or actually lack of routes) to `peer5`. So peers know about `peer5` because of archipelago/bff right?

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+```typescript
+
+type Address = string
+
+// A route is a list of peer ids. It doesn't contain the source or destination, so for example if a->b->c
+// the route between a and c is [ b ]
+// and the route between b and c is []
+type Route = Address[]
+
+// A map between peer and a route. If a peer is missing, it means there is no direct or indirect P2P connection to the peer, so the messaging server should be used
```
marianogoldman, line 170:
```suggestion
// A map between peer and a route. If a peer is missing, it means there is no direct or indirect P2P
// connection to the peer, so the messaging server should be used
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+type PeerRoutingTable = Map<Address, Route>
+```
+
+## Messaging service
+
+To establish a P2P webrtc connection, peers will exchange signals with each other using the messaging service. This service will act as a fallback when no P2P route is available to deliver a message.
```
marianogoldman, line 176: So... the messaging service has the responsibility of doing the signalling?

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+
+This service has same level of trust than any peer in the network, this means it has no specific authentication requirements, and any trust feature has to be built in the message itself. That been said implementators may choose to add some extra validation when doing high level clustering (like [real-time clustering of users in Islands (ADR-35)](/adr/ADR-35)), for clusters to remain isolated.
```
marianogoldman, line 178:
```suggestion
This service has same level of trust than any peer in the network, this means it has no specific authentication requirements, and any trust feature has to be built in the message itself. That been said implementors may choose to add some extra validation when doing high level clustering (like [real-time clustering of users into Islands (ADR-35)](/adr/ADR-35)), for clusters to remain isolated.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+          if (route === 'server' || route.length > relayedRoute.length) {
+            route = relayedRoute
+          }
+        }
+      }
+    }
+
+    calculatedRoutes.set(toPeer, route)
+
+    // NOTE: routes are bidirectional
```
marianogoldman, line 274: Perhaps we should mention this above in the text somewhere.

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+    }
+  }
+
+  return routingTables
+}
+```
+
+## Packet
+
+A packet contains a source (peer id) and map specifying to whom the packet is for and the route to follow to reach it.
```
marianogoldman, line 304:
```suggestion
A packet contains a source (peer id) and map specifying who the packet is for and the route to follow to reach it.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+```typescript
+{ 
+  source: 'peer1', 
+  to: { 'peer2': [], 'peer3': ['peer2'] }
+}
+```
+
+This means each peer will check the `to` field, if they are one of the recipent of the messages the message will be processed, otherwise it will be relayed in the way the route value indicates.
+
+Notice the peer relaying the message is not expected to remove itself either as a recipent or as a hop in the route, since this will require encoding the package again.
```
marianogoldman, line 327: So, let's suppose the intermediate peer has a connection to the peer from which it received the message... it has to be clever enough to parse and pick up from anything after itself, right?

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+This means each peer will check the `to` field, if they are one of the recipent of the messages the message will be processed, otherwise it will be relayed in the way the route value indicates.
+
+Notice the peer relaying the message is not expected to remove itself either as a recipent or as a hop in the route, since this will require encoding the package again. 
+
+# Benefits
+
+- Since there is a specific routing table for each message, there is no need to expire message or count hops. If a route is cut, the message will not be delivered by the mesh, this means the message should be relayed using the messaging service (if reliable) or discarded (if unreliable).
+- The messaging service fallback provide a safety guarantee against network clusters.
```
marianogoldman, line 332:
```suggestion
- The messaging service fallback provide a safety guarantee against network cluster partitions.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+- A given implementation can be optimized by suggesting peers to connect to certain others in order to avoid clustering and minimize messaging service usage. This is out of the scope for this document.
+- Since the routing service will know the status of the mesh at all times, it's easy to graph and debug network problems. 
+
+# Competition (alternatives)
+
+- One simple solution using a P2P mesh, in which peers broadcast to others, and upon receiving a message, they relay to their connections. The problem with this solution is there is the possibility of messages going thought the network after the same message is deliver, which implies the need to build a mechanism to discard old messages (using time or hops). Another problem is the possibility of clusters in the mesh, which may cause certain group of peers, to not see the ones in another group.
```
marianogoldman, line 338:
```suggestion
- One simple solution using a P2P mesh, in which peers broadcast to others, and upon receiving a message, they relay to their connections. The problem with this solution is there is the possibility of messages going thought the network after the same message is delivered, which implies the need to build a mechanism to discard old messages (using time or hops). Another problem is the possibility of clusters in the mesh, which may cause certain group of peers to not see the ones in another group (partitioned).
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+
+# Non-goals
+
+- The route representation is not as efficient as it could be, since uses Address (string) as a peer identification. This should be possible to improve, by assiging numeric alias for every peer. However, detailing how to do that is out of the scope of this document, and doesn't change the overall proposed solution.
```
marianogoldman, line 342:
```suggestion
- The route representation is not as efficient as it could be, since it uses `Address` (string) as a peer identification. This should be possible to improve, by assigning a numeric alias to every peer. However, detailing how to do that is out of the scope of this document, and doesn't change the overall proposed solution.
```

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+
+# Key dependencies and Open questions
+
+- Why not let each peer calculate the best routing table, by broadcasting each peer status to the network?.
```
marianogoldman, line 346:
```suggestion
- Why not let each peer calculate the best routing table, by broadcasting each peer status to the network?
```

RFC/RFC-14-p2p-adapter.md
hugoArregui: yes, only if in the path to other, so say you have the following mesh:

`a<->b<->c` and `a<->d`

a will send the following message:


```typescript
{
  source: 'a'
  to: {
  'b': [],
  'c': ['b'],
  'd': []
  }
}
```


d will receive it, parse it, and process it
c will receive it, parse it, and process it. But attempt to relay to c as well, because is the route to d. If the connection is cut, it will try to reach d thought the messaging service.

I tried to illustrate this on the peer is disconnected example, but let me know if you think I'm missing something

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+  - agusAldasoro
+---
+
+## Abstract
+
+This document describes a possible implementation for a P2P adapter using a fallback messaging service. For more information about what an adapter is, check [ADR-81](/ADR/ADR-81-minimum-comms-transport.md).
+
+## Need
+
+A P2P adapter is needed in order to provide a light mechanism for comms (compared to [RFC-5: comms service using websocket](/RFC/RFC-5-ws-comms.md)). This adapter should guarantee the delivery of a message to any peer in the network (or cluster, [real-time clustering of users in Islands (ADR-35)](/adr/ADR-35)).
```
menduz, line 18: What real-world problem is this RFC trying to address? It seems like a P2P+TURN server mesh should be compliant with all the things mentioned here, and a way less complex solution to the described problem. Or maybe I don't yet see the key point of this proposal

RFC/RFC-14-p2p-adapter.md
```diff
@@ -0,0 +1,348 @@
...
+- The messaging service fallback provide a safety guarantee against network cluster partitions.
+- A given implementation can be optimized by suggesting peers to connect to certain others in order to avoid clustering and minimize messaging service usage. This is out of the scope for this document.
+- Since the routing service will know the status of the mesh at all times, it's easy to graph and debug network problems. 
+
+# Competition (alternatives)
+
+- One simple solution using a P2P mesh, in which peers broadcast to others, and upon receiving a message, they relay to their connections. The problem with this solution is there is the possibility of messages going thought the network after the same message is delivered, which implies the need to build a mechanism to discard old messages (using time or hops). Another problem is the possibility of clusters in the mesh, which may cause certain group of peers to not see the ones in another group (partitioned).
```
menduz, line 338:
```suggestion
- One simple solution using a P2P mesh, in which peers broadcast to others, and upon receiving a message, they relay to their connections. The problem with this solution is there is the possibility of messages going thought the network after the same message is delivered, which implies the need to build a mechanism to discard old messages (using time or hops). Another problem is the possibility of clusters in the mesh, which may cause certain group of peers to not see the ones in another group (partitioned).
- Keep the Routing Service to help building the Mesh, reducing the logic in peers and enabling lightweight ports of P2P comms to other technologies (using the same backend and mesh algorithms). Also use TURN to reduce the moving parts instead of the MessagingService
```

menduz (commented):
I think that adding a MessageService defeats the purpose of a lightweight protocol, instead I'd recommend to rely on TURN and P2P mesh.

For the routing service, it seems like we are proposing to re-implement BGP, I'd recommend to take a look at that protocol it is the protocol that literally keeps the world running.

Another suggestion is to unify the p2p connection management (candidates, offers, etc) with this routing protocol, and move that logic to this RoutingService. That way, the server can have full context of the mesh at all times (removing the need of reporting) and optimize the network accordingly. There is more: implementing vanilla libwebrtc based comms in native clients will be trivial, since the logic to pick the candidates will be entirely server-based. Greatly benefiting "the protocol" and the implementer's lives. Along with consistent fixes applied to servers without needing to redistribute the native clients to all devices and across all app-stores.

RFC/RFC-14-p2p-adapter.md
hugoArregui: > I think that adding a MessageService defeats the purpose of a lightweight protocol, instead I'd recommend to rely on TURN and P2P mesh.

Well, the message service is a fallback for when cluster happens in the network, the idea of this proposal is to have a algorithm that's simple, have a lot of visibility in what's happening in the network and is subject to improvements. Selecting random candidates may end up reducing the probably of forming internal clusters and thus the message service won't be used. But we have a guarantee that even if clusters appear, we will deal with the situation. Furthermore, if we have clusters and we decide that's a problem, we can suggest new ideas (probably different implementations on top of this RFC) for the server to suggest connections in order to bridge and unify cluster or any others. 

> 
> For the routing service, it seems like we are proposing to re-implement BGP, I'd recommend to take a look at that protocol it is the protocol that literally keeps the world running.

I think we are probably implementing BGP, yes. If there is a library out there we can use it.
 
> Another suggestion is to unify the p2p connection management (candidates, offers, etc) with this routing protocol, and move that logic to this RoutingService. That way, the server can have full context of the mesh at all times (removing the need of reporting) and optimize the network accordingly. There is more: implementing vanilla libwebrtc based comms in native clients will be trivial, since the logic to pick the candidates will be entirely server-based. Greatly benefiting "the protocol" and the implementer's lives. Along with consistent fixes applied to servers without needing to redistribute the native clients to all devices and across all app-stores.

As stated above the messaing service is mostly a fallback, if the network has no clusters and it's stable, the service won't be used.

RFC/RFC-14-p2p-adapter.md
hugoArregui: (sorry I missed this comment): yes, it does, right now the topics service is doing that but we could leverage this new service, which is more precise

RFC/RFC-14-p2p-adapter.md
hugoArregui: yes

RFC/RFC-14-p2p-adapter.md
hugoArregui: We can continue this in another PR, I'm resolving the conversation in order to merge this one.

RFC/RFC-14-p2p-adapter.md
hugoArregui: We can continue this in another PR, I'm resolving the conversation in order to merge this one.

PR #109, opened 2022-10-27
feat: Feat/p2p updates

agusaldasoro:


RFC/RFC-14-p2p-adapter.md
```diff
@@ -90,70 +91,79 @@ sequenceDiagram
     participant Peer5
     participant RS as Routing Service
 
-    RS->>Peer1: { peer2: [], peer3: [], peer4: [] }
-    RS->>Peer2: { peer1: [], peer3: [], peer4: [ peer1 ] }
-    RS->>Peer3: { peer1: [], peer2: [], peer4: [ peer1 ] }
-    RS->>Peer4: { peer1: [], peer2: [ peer1 ], peer3: [ peer1 ] }
-    RS->>Peer5: {}
+    RS->>Peer1: { paths: [ [1, 2], [1, 3], [1, 4] ], unreachable: [5] }
```
hugoArregui, line 94: I don't think the routing service should be explicit about the unreachable routes, since the peer knows every peer, then it can deduce from the peers there is no route to, they are unrecheable

RFC/RFC-14-p2p-adapter.md
```diff
@@ -162,13 +172,11 @@ sequenceDiagram
...
-// A route is a list of peer ids. It doesn't contain the source or destination, so for example if a->b->c
-// the route between a and c is [ b ]
-// and the route between b and c is []
+// A route is a list of peer ids.
 type Route = Address[]
 
-// A map between peer and a route. If a peer is missing, it means there is no direct or indirect P2P connection to the peer, so the messaging server should be used
-type PeerRoutingTable = Map<Address, Route>
+// A list of all the paths that need to be covered to broadcast the network
+type PeerRoutingTable = List<Route>
```
hugoArregui, line 179: Route[]?

RFC/RFC-14-p2p-adapter.md
```diff
@@ -297,32 +307,32 @@ function calculateRoutingTables(mesh: Map<string, Set<string>>) {
...
   source: Address
-  to: Record<Address, Route>
+  target: List<Route>
 }
 ```
 
 ## Peer
 
-The peer will send a message to every known peer using the routing table, sending it once to each peer. Let's say peer1 is connected to peer2 and peer2 to peer3, when broadcasting to the network peer1 will send:
+Each time a package needs to be send, then the peer will create it using its own paths as routing table and will make the network to obay that flow.
```
hugoArregui, line 326: s/obay/obey

RFC/RFC-14-p2p-adapter.md
```diff
@@ -297,32 +307,32 @@ function calculateRoutingTables(mesh: Map<string, Set<string>>) {
...
 
-A packet contains a source (peer id) and map specifying to whom the packet is for and the route to follow to reach it.
+A packet contains a source (peer id) and a target specifying to whom the packet is for and the route to follow to reach it.
 
 ```typescript
 
 type Packet = {
   source: Address
-  to: Record<Address, Route>
+  target: List<Route>
```
hugoArregui, line 320: Route[]

RFC/RFC-14-p2p-adapter.md
```diff
@@ -221,6 +229,8 @@ type NewPeerRoutingTableEvent = {
 
 Reference implementation:
 
+// TODO: Update this
```
hugoArregui, line 232: I don't think we need to update it right now, since the algorithm has the guarantees we need and it doesn't say anything regarding how the package will be encoded

RFC/RFC-14-p2p-adapter.md
agusaldasoro: What is the change suggestion here?

RFC/RFC-14-p2p-adapter.md
```diff
@@ -297,32 +368,31 @@ function calculateRoutingTables(mesh: Map<string, Set<string>>) {
...
   source: Address
-  to: Record<Address, Route>
+  target: Route[]
 }
-```
 
 ## Peer
 
-The peer will send a message to every known peer using the routing table, sending it once to each peer. Let's say peer1 is connected to peer2 and peer2 to peer3, when broadcasting to the network peer1 will send:
+Each time a package needs to be send, then the peer will create it using its own paths as routing table and will make the network to obay that flow.
```
agusaldasoro, line 386:
```suggestion
Each time a package needs to be sent, then the peer will create it using its own paths as routing table and will make the network to obey that flow.
```

PR #143, opened 2022-11-28
RFC 143: Social service authentication

jmoguilevsky:


content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+## Abstract
+
+<!--
+Insert here a brief paragraph summarizing the RFC in its current state.
+This section shall provide an overview of whether this is a settled
+decision, alternatives explored and a short summary of relevant
+background information and key insights.
+-->
+
+As part of the social service initiative users will need to be able to authenticate themselves to be able to interact with other users and be able to access their information correctly.
```
menduz, line 21: The abstract does not describe the current status of the document nor summarizes the content of it.

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+The endpoint that we have chosen is `_matrix/client/r0/pushrules/` but could change if we find one that suits better.
+
+## Benefit
+
+This solution has two main benefits:
+
+1. There's no need to create a new authorization/authentication service nor system
+1. The solution is completely seamless for the user and doesn't imply no persistance of tokens from the social service side
+
+## Competition
```
menduz, line 90: prior art ADR-44

menduz (changes requested):
I requested some changes for clarity of content

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
+---
+adr: 138
```
marianogoldman, line 2: Please use ADR id same as PR (in this case 143) to avoid conflicts with other ADRs (there's already another one ongoing with same id).

content/ADR-138-social-service-authentication.md
jmoguilevsky: That would be a new authentication system, until we have that, we can continue leveraging Matrix's one

menduz: We authenticate the chat using ethereum signatures nowadays. That is what we send to matrix. Did you consider using ethereum signatures to authenticate against this service?

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+<!--
+What other options were considered? Give an honest treatment of why
+these alternatives were not satisfactory. Identify the competition and
+demonstrate that the competition is clearly understood. Include the
+“what if we do nothing” alternative.
+-->
+
+### Implementing an authentication service
+
+This is a great option for when we want to stop using Matrix, we would be in control of everything, allowing to build login features as we deem necessary. The biggest problem of this option is the cost
```
LautaroPetaccio, line 101: Just a small suggestion:
```suggestion
This is a great of if we ever want to stop using Matrix. We would be in control of everything, allowing to build login features as we deem necessary. Although this brings a lot of flexibility on login features, the biggest problem of this option is the cost of implementing everything from scratch, which is not necessary due to the current requirements.
```

LautaroPetaccio (commented):
I've just left some suggestions, nothing mandatory, it's just to see if you like them.

content/ADR-138-social-service-authentication.md
kevinszuchet, line 21: Small suggestion:
```suggestion
As part of the social service initiative, users will need to authenticate themselves to interact with other users and to access their information correctly.
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+doing and the rationale behind the decisions you propose.  Then lay out
+the plan of execution, in rough order of how the execution should take
+place.  Include the rollout plan as well. (This is usually the longest
+section of the RFC) Hint: don’t be afraid of posting illustrations! The
+level of detail here has to be enough to give the reader  a clear
+understanding of the solution - it is up to the writer to decide.
+Further detail can be addressed to satisfy comments and increase clarity.
+-->
+
+The logging in system for the social needs to solve three problems:
```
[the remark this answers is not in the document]
kevinszuchet: I would change also:
```suggestion
The logging-in system for the social service needs to solve three problems:
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. The user is correctly authenticated (the token is valid)
+1. Who the user is (which userId does the token belong to)
+1. Does the user have the correct permissions to perform a certain action (Authorization)
+
+The first two problems are already solved by the current chat solution (Matrix), so we propose to leverage the already existing implementation for simplicity and velocity in development. Matrix will be used as an authentication system, users will login against their service, then each endpoint will first validate against Matrix that the user is still using a valid token.
```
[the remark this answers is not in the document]
kevinszuchet: Punctuation and conciseness suggestions:

```suggestion
The first two problems are already solved by the current chat solution (Matrix), so we propose to leverage the existing implementation for simplicity and velocity in development. Matrix will be used as an authentication system where users will log in against their service. Then, each endpoint will validate against Matrix that the user is still using a valid token.
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+
+The service will have a distributed cache (without persistance) with user-to-token information which will be used both to obtain the user (once the token is validated) and to prevent hitting Matrix for every request.
```
kevinszuchet, line 44: Small suggestion:
```suggestion
The service will have a distributed cache (without persistence) with user-to-token information, which will be used to obtain the user (once the token is validated) and to prevent hitting Matrix for every request.
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+    User -->|Request| Social-service(Social service)
+    Social-service -->|Get User | Cache
+    Cache -->|User| Social-service
+    Social-service --> C{User present}
+
+    C -->|True| D(Continue request)
+    C -->|False| Validate-token(Validate token against matrix)
+    Validate-token --> |Get User| Matrix
+    Matrix --> |User| Social-service
+```
```
kevinszuchet, line 70: When you interact with the `Cache` and `Matrix`, are you getting the User or the Token?

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+
+In this second workflow the token life in the cache will be extended by what's defined in matrix, thus preventing the user from using an invalid token.
```
kevinszuchet, line 72: Small suggestion:
```suggestion
In this second workflow, the token life in the cache will be extended by what's defined in Matrix, thus preventing the user from using an invalid token.
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+If the token is expired and the user tries to use it again, the service will validate against Matrix, if it's still valid, the user will be obtained again and re-stored in the Cache.
```
kevinszuchet, line 73: Small suggestion:
```suggestion
If the token is expired and the user tries to use it again, the service will validate against Matrix, and if it's still valid, the user will be obtained again and re-stored in the Cache.
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+
+To validate the token against Matrix, an authenticated endpoint must be chosen. The endpoint needs to satisfy the following specifications:
+
+- No to low impact on the database
+- No to low usage from the client (to prevent overloading it even more)
+- Ideally tackled by a specific worker so in the case it's necessary it would be easier to scale up
```
kevinszuchet, line 79: Small punctuation suggestion:
```suggestion
Ideally tackled by a specific worker, so in case it's necessary, it would be easier to scale up
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+
+The endpoint that we have chosen is `_matrix/client/r0/pushrules/` but could change if we find one that suits better.
+
+## Benefit
+
+This solution has two main benefits:
+
+1. There's no need to create a new authorization/authentication service nor system
```
kevinszuchet, line 87:
```suggestion
1. There's no need to create a new authorization/authentication service or system
```

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+1. The solution is completely seamless for the user and doesn't imply no persistance of tokens from the social service side
```
kevinszuchet, line 88:
```suggestion
1. The solution is completely seamless for the user and doesn't imply the persistence of tokens from the social service side
```

content/ADR-138-social-service-authentication.md
jmoguilevsky: The signature is sent as login system it replaces the password, but each request is authenticated with a JWT built by synapse

jmoguilevsky: In this RFC I don't imply changing the logging in system, nor the authentication model, only how we handle the token in the social service

content/ADR-138-social-service-authentication.md
jmoguilevsky: From the cache you obtain the user.

When you interact with Matrix it depends, which interaction are you asking about?

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+<!--
+What other options were considered? Give an honest treatment of why
+these alternatives were not satisfactory. Identify the competition and
+demonstrate that the competition is clearly understood. Include the
+“what if we do nothing” alternative.
+-->
+
+### Implementing an authentication service
+
+This is a great of if we ever want to stop using Matrix. We would be in control of everything, allowing to build login features as we deem necessary. Although this brings a lot of flexibility on login features, the biggest problem of this option is the cost of implementing everything from scratch, which is not necessary due to the current requirements.
```
LautaroPetaccio, line 101: Sorry, I made a mistake when I made the suggestion 🤦 
```suggestion
This is a great option if we ever want to stop using Matrix. We would be in control of everything, allowing to build login features as we deem necessary. Although this brings a lot of flexibility on login features, the biggest problem of this option is the cost of implementing everything from scratch, which is not necessary due to the current requirements.
```

content/ADR-138-social-service-authentication.md
kevinszuchet: My bad. I think I understand now. After logging in, you have the token to get the user from the Cache or Matrix. You have one cache for tokens and another for users, right?

content/ADR-138-social-service-authentication.md
```diff
@@ -0,0 +1,101 @@
...
+In this second workflow the token life in the cache will be extended by what's defined in matrix, thus preventing the user from using an invalid token.
+If the token is expired and the user tries to use it again, the service will validate against Matrix, if it's still valid, the user will be obtained again and re-stored in the Cache.
+
+To validate the token against Matrix, an authenticated endpoint must be chosen. The endpoint needs to satisfy the following specifications:
+
+- No to low impact on the database
+- No to low usage from the client (to prevent overloading it even more)
+- Ideally tackled by a specific worker so in the case it's necessary it would be easier to scale up
+
+The endpoint that we have chosen is `_matrix/client/r0/pushrules/` but could change if we find one that suits better.
```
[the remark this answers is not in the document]
jmoguilevsky: The documentation doesn't specify which kind of token, it says it's an access token, for the structure it looks like a JWT, but can't be sure https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-login

content/ADR-138-social-service-authentication.md
jmoguilevsky: The Cache has `token->userInfo` so with the token I can get the user.

When the user logs in, we save the register `token: user` in the Cache, in every other request we obtain the user using the token.

If the token is not in the cache, we go with the token to matrix to obtain the user, and then if it's a successful request we save the same register in the cache.  Thus, there's need for only one cache.

content/ADR-138-social-service-authentication.md
jmoguilevsky: Yep, I already fixed it in a commit

content/ADR-138-social-service-authentication.md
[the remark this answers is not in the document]
jmoguilevsky: Specified

content/ADR-138-social-service-authentication.md
jmoguilevsky: Updated the abstract

content/ADR-138-social-service-authentication.md
menduz: I think that the fact of "handling tokens" already assumes many things of the authentication model, those are worth discussing before moving on

menduz: Oh, the new abstract is SO good. I think we can resolve this

menduz (approved):

PR #146, opened 2022-12-01
cancel friend request changes and replaced GetFriendRequestsV2 & AddFriendRequestsV2

lorux0:
Signed-off-by: Nicolas Lorusso <56365551+lorux0@users.noreply.github.com>

### DIFF
--- a/content/ADR-137-new-friend-request.md
+++ b/content/ADR-137-new-friend-request.md
@@ -161,7 +161,7 @@ acceptFriendshipPayload {
 ```
 acceptFriendshipConfirmationPayload{
     messageId: string,
-    friendRequestId: string
+    friendRequest: friendRequestPayload
 }
 ```
 
@@ -199,7 +199,7 @@ rejectFriendshipPayload {
 ```
 rejectFriendshipConfirmationPayload{
     messageId: string,
-    friendRequestId: string
+    friendRequest: friendRequestPayload
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sandrade-dcl (approved):

content/ADR-137-new-friend-request.md
```diff
@@ -104,7 +104,7 @@ cancelFriendshipPayload {
 ```
 cancelFriendshipConfirmationPayload{
     messageId: string,
-    friendRequestId: string
+    friendRequest: friendRequestPayload
```
Julieta11, line 107: It's not a problem, but I'm wondering: why would you need us to send you the `friendRequestPayload` instead of the `friendRequestId` when the user cancels a friend request?

content/ADR-137-new-friend-request.md
```diff
@@ -217,19 +217,19 @@ sequenceDiagram
   participant renderer
   participant kernel
   
-renderer->kernel: GetFriendRequests(getFriendRequestsPayload{sentLimit:50, sentSkip:0, receivedLimit:50, receivedSkip:0})
+renderer->kernel: GetFriendRequestsV2(getFriendRequestsPayload{sentLimit:50, sentSkip:0, receivedLimit:50, receivedSkip:0})
```
Julieta11, line 220: I think the `messageId` param is missing in `getFriendRequestsPayload`, isn't it?

content/ADR-137-new-friend-request.md
lorux0: we need to get the user related to the request, we dont have that kind of information in the context that we are receiving the message

marianogoldman (approved):

PR #157, opened 2022-12-20
add more clarity about the pixel density and layouting decisions

menduz:
Adds clarifications about pixel density and how it affects the UiTransform. TLDR: mimics the behavior of the browsers where a `1(device)px*devicePixelRatio=1(sdk)px`. Introduces the UiCanvasInformation to retrieve sizes and ratios from the SDK


content/ADR-124-flexbox-based-ui-for-scenes.md
```diff
@@ -31,18 +31,24 @@ The renderer systems must ignore in the 3D camera every entity containing a `UiT
...
+The `RootEntity` will act as "root node" and will not accept a `UiTransform` component, its size will be equivalent to the full viewport of the Rendering engine. Enabling its children to occupy "100%" width and height, as well relative and absolute positioning in the corners.
+
+Given the difficulty of creating UIs with custom viewport and pixel density sizes present on the SDK6, the `UiTransform` will represent pixels scaled by the current pixel density of the device screen (`devicePixelRatio`). Meaning one screen pixel will be one UiTransform pixel on a pixel density of 1 (low DPI display) and it will be adjusted to the current configuration in high DPI displays like the ones present in modern laptops (e.g. retina display). This should not limit the capabilities of the SDK to create UIs that are reactive to the size of the screens, because percentages are available as units of measurement and the new `UiCanvasInformation` fills in the gaps of information. This behavior mimics the `devicePixelRatio` of web browsers, in which a pixel is always represented as a pixel in CSS, besides being adjusted to 2(device)pixels in a retina display (`devicePixelRatio=2`).
```
marianogoldman, line 36:
```suggestion
Given the difficulty of creating UIs with custom viewport and pixel density sizes present on the SDK6, the `UiTransform` will represent pixels scaled by the current pixel density of the device screen (`devicePixelRatio`). Meaning one screen pixel will be one `UiTransform` pixel on a pixel density of 1 (low DPI display) and it will be adjusted to the current configuration in high DPI displays like the ones present in modern laptops (e.g. retina display). This should not limit the capabilities of the SDK to create UIs that are reactive to the size of the screens, because percentages are available as units of measurement and the new `UiCanvasInformation` fills in the gaps of information. This behavior mimics the `devicePixelRatio` of web browsers, in which a pixel is always represented as a pixel in CSS, besides being adjusted to 2(device)pixels in a retina display (`devicePixelRatio=2`).
```

content/ADR-124-flexbox-based-ui-for-scenes.md
```diff
@@ -31,18 +31,24 @@ The renderer systems must ignore in the 3D camera every entity containing a `UiT
...
+
+`UiCanvasInformation` serves a purpose of providing information to the secene for custom layouting based on `devicePixelRatio`, device orientation and canvas sizes are in exclusive charge of the implementation of the SDK libraries themselves. Removing any layouting responsibility from the renderer for the sake of keeping the semantics simple and backwards compatible. This will create an inversion of control in which the scene will decide everything in relation to the scene UI.
```
marianogoldman, line 38:
```suggestion
`UiCanvasInformation` serves a purpose of providing information to the secene for custom laying out based on `devicePixelRatio`, device orientation and canvas sizes are in exclusive charge of the implementation of the SDK libraries themselves. Removing any laying out responsibility from the renderer for the sake of keeping the semantics simple and backwards compatible. This will create an inversion of control in which the scene will decide everything in relation to the scene UI.
```

content/ADR-124-flexbox-based-ui-for-scenes.md
```diff
@@ -237,6 +243,23 @@ message PBUiTransform {
...
+
+```protobuf
+message UiCanvasInformation {
+  // informs the scene about the resolution used for the UI rendering
+  float device_pixel_ratio = 1;
+  // informs about the width of the canvas
+  int32 width = 2;
+  // informs about the height of the canvas
+  int32 height = 3;
+  // informs the sdk about the interactable area. some implementations may change this area depending on the HUD that is being shown. this value may change at any time by the Renderer to create reactive UIs
```
marianogoldman, line 258:
```suggestion
  // informs the sdk about the interactable area. Some implementations may change this area depending on the HUD that is being shown. This value may change at any time by the Renderer to create reactive UIs.
```

marianogoldman (approved):

content/ADR-124-flexbox-based-ui-for-scenes.md
menduz: I meant layouting there, referring to the process of calculating the layout and not the layout itself

mikhail-dcl:
It can be achieved in Unity by setting "Constant Pixel Size" mode and adjusting `Scale` parameter.
image
Despite there is no such concept as "devicePixelRatio" we can introduce it and calculate from the code.
At this point it looks adequately doable.
I just want to warn you that we will not be able to achieve 100% fidelity with "browser" capabilities, and in pursuing it we will be spending much more resources than normally required. We should take into account and Unity's UI Toolkit capabilities and the design flow it provides as it is our main renderer maintained by us.
For now it's good to go.

mikhail-dcl (approved):

pravusjif (approved):

content/ADR-124-flexbox-based-ui-for-scenes.md
marianogoldman: Layouting doesn't exist (I believe). Not formal English at least.
https://forum.wordreference.com/threads/laying-out-or-layouting.3079156/

PR #162, opened 2022-12-28
feat: extend auth chain to include permissions

marianogoldman:


content/ADR-162-auth-chain-with-permissions.md
```diff
@@ -0,0 +1,311 @@
...
+An alternative approach to permission specification would be to use structured
+data signing, as defined in
+[EIP-712 Typed structured data hashing and signing](https://eips.ethereum.org/EIPS/eip-712).
+
+The follow snippet can be run in the browser console:
+
+```javascript
+let domain = [
+    {name: "name", type: "string"},
+    {name: "chainId", type: "uint256"}
```
nachomazzara, line 209: chainId is not mandatory. If we pass the `chainId` the users won't be able to sign the message if they are in a different network. This is the same case when user signs meta transactions for Polygon by being connected to Ethereum. We use the [`salt` parameter to specify the chainId](https://github.com/decentraland/wearables-contracts/blob/master/contracts/commons/EIP712Base.sol#L16). 

I may suggest adding the version here as well.

About the salt, we can use it if we want to prevent ambiguity for some reason

marianogoldman: Thanks Nacho. This snipped of code came from some example I got on the web, we definitely need to revise the details, so whatever you guys suggest is fine.
This doc has a lot to be updated. The main part that needs revision is the way we express the permissions so that it is valid for the use cases we have today. And apart from being usable from backend side, it also has to be decently presented for end users who have to sign the ephemeral key. 
@fzavalia was going to try out a few ways to organize the permission data to see if we can decide on a way that covers both sides (clear to the user that's signing the key and expressive enough for common scenarios in the explorer or other websites).

menduz:
Merging to not lose track of this

content/ADR-162-auth-chain-with-permissions.md
```diff
@@ -0,0 +1,311 @@
...
+    deny: "dcl:explorer:denied-operation"
+```
+
+The `resource` name must be an item referencing an entity of the domain that
+the server knows. For e.g. for DCL Worlds, it can be a DCL name, while for
+scene deployment in Genesis City it can be the coordinates of the land, or
+for a profile deployment it can be the profile address.
+
+The `allow` and `deny` elements are structured as follows:
+`namespace:service-name:operation-name` or `namespace:service-name:*`.
```
guidota, line 136: Could those elements be anything or would it be a list of known operations? 

Thinking in bad actors and the case of _denying_ misspelled operations and users thinking it is OK. For example,
let's say we have the operation `dcl:explorer:deploy` but some signs this:
```  
- resource: "0xaddress"
  allow: "dcl:explorer:*"
- resource: "0xaddress"
  deny: "dcl:explorer:deployment"
```
but it actually gets `dcl:explorer:deploy` operation allowed. Could this be a problem?

Also, by `least privilege` I understand that all operations would be _denied_ by default, is that right?

marianogoldman:
We should merge it not to lose what was already discussed, but we never followed up on this one.
@menduz please approve or merge directly

content/ADR-162-auth-chain-with-permissions.md
marianogoldman: It's kind of stale right now, there are a few things that already need re-thinking.
Happy to open up again for discussion / work on it.

menduz:
Roger

PR #168, opened 2023-01-09
Entity definition and GSet addition to CRDT

leanmendoza:
Signed-off-by: Lean Mendoza <leandro@decentraland.org>

### DIFF
--- a/content/ADR-166-entities-and-crdt-modification.md
+++ b/content/ADR-166-entities-and-crdt-modification.md
@@ -9,6 +9,9 @@ authors:
   - leanmendoza
 ---
 
+## Abstract
+This ADR introduces the last definition of an Entity for SDK7 and also aims to change the ADR-117 CRDT specification adding a rule and a command.
+
 ## Context
 
 With the precedent of the ADRs (quote the adrs), the bases for the type of data and the basic rules of consensus between peers were defined. In the framework of the ECS system, the entity is a representative number within a data collection of a certain component and this number was established as a 32-bit positive integer. Up to now, no limit has been established (neither theoretical nor practical mention) on the number of supported entities, but it is inevitable to think that 2^32 of entities is an insane number.
--- a/content/ADR-166-entities-and-crdt-modification.md
+++ b/content/ADR-166-entities-and-crdt-modification.md
@@ -1,0 +1,95 @@
+---
+adr: 123
+title: CRDT Entity spaces
+date: 2022-05-09
+status: Living
+type: Standards Track
+spdx-license: CC0-1.0
+authors:
+  - leanmendoza
+---
+
+## Context
+
+With the precedent of the ADRs (quote the adrs), the bases for the type of data and the basic rules of consensus between peers were defined. In the framework of the ECS system, the entity is a representative number within a data collection of a certain component and this number was established as a 32-bit positive integer. Up to now, no limit has been established (neither theoretical nor practical mention) on the number of supported entities, but it is inevitable to think that 2^32 of entities is an insane number.
+
+Over the course of a scene, entities can be created and destroyed. Today, the destruction of entities involves their deprecation, since each component will have its history (timestamp > 0), even though it implies a null final state (no data in all components). It is important to highlight it so that there is no conflict in the definition of its state. For example: for the parenting of entities we use the Transform component, and suppose that in entity B we point to entity A. If we destroy entity A, the default behavior will be that entity B re-points to the root entity ( if it was not made explicit that the children of A should be destroyed). If we wanted to reuse the entity A (the number) with another "meaning", it would have an unwanted side effect, since A meant something else in the past.
+
+On the other hand, if we didn't reuse these values but wanted to preserve the state for conflict resolution, for example, an update on a deleted entity, we would have as many states for entities as historically created entities, regardless of the limit of net entities in the same instant.
+
+In summary, both situations present their problems:
+  - A) reusing entities presents side-effects in the use and potential problems in the resolution of conflicts
+  - B) not reusing entities supposes a memory leak where dead states would be stored
+
+So what needs to be solved is:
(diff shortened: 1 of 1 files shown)


content/ADR-166-entities-and-crdt-modification.md
```diff
@@ -0,0 +1,98 @@
...
+authors:
+  - leanmendoza
+---
+
+## Abstract
+This ADR introduces the last definition of an Entity for SDK7 and also aims to change the ADR-117 CRDT specification adding a rule and a command.
+
+## Context
+
+With the precedent of the ADRs (quote the adrs), the bases for the type of data and the basic rules of consensus between peers were defined. In the framework of the ECS system, the entity is a representative number within a data collection of a certain component and this number was established as a 32-bit positive integer. Up to now, no limit has been established (neither theoretical nor practical mention) on the number of supported entities, but it is inevitable to think that 2^32 of entities is an insane number.
```
marianogoldman, line 17:
```suggestion
With the precedent of the ADRs (quote the adrs), the basis for the type of data and the basic rules of consensus between peers were defined. In the framework of the ECS system, the entity is a representative number within a data collection of a certain component and this number was established as a 32-bit positive integer. Up to now, no limit has been established (neither theoretical nor practical mention) on the number of supported entities, but it is inevitable to think that 2^32 of entities is an insanely big number.
```

content/ADR-166-entities-and-crdt-modification.md
```diff
@@ -0,0 +1,98 @@
...
+
+Over the course of a scene, entities can be created and destroyed. Today, the destruction of entities involves their deprecation, since each component will have its history (timestamp > 0), even though it implies a null final state (no data in all components). It is important to highlight it so that there is no conflict in the definition of its state. For example: for the parenting of entities we use the Transform component, and suppose that in entity B we point to entity A. If we destroy entity A, the default behavior will be that entity B re-points to the root entity ( if it was not made explicit that the children of A should be destroyed). If we wanted to reuse the entity A (the number) with another "meaning", it would have an unwanted side effect, since A meant something else in the past.
```
marianogoldman, line 19:
```suggestion
Over the course of a scene, entities can be created and destroyed. Today, the destruction of entities involves their deprecation, since each component will have its history (timestamp > 0), even though it implies a `null` final state (no data in all components). It is important to highlight it so that there is no conflict in the definition of its state. For example: for the parenting of entities we use the `Transform` component, and suppose that in entity B we point to entity A. If we destroy entity A, the default behavior will be that entity B re-points to the root entity (if it was not made explicit that the children of A should be destroyed). If we wanted to reuse the entity A (the number) with another "meaning", it would have an unwanted side effect, since A meant something else in the past.
```

marianogoldman, line 19: Not 100% percent sure of this... but "suppose that in entity B we point to entity A"... shouldn't it be the other way around? i.e. "suppose that in entity A we point to entity B"?

content/ADR-166-entities-and-crdt-modification.md
```diff
@@ -0,0 +1,98 @@
...
+
+On the other hand, if we didn't reuse these values but wanted to preserve the state for conflict resolution, for example, an update on a deleted entity, we would have as many states for entities as historically created entities, regardless of the limit of net entities in the same instant.
+
+In summary, both situations present their problems:
+  - A) reusing entities presents side-effects in the use and potential problems in the resolution of conflicts
+  - B) not reusing entities supposes a memory leak where dead states would be stored
+
+So what needs to be solved is:
+- A) The entities MUST be unequivocal, and therefore their numbers: unique identifiers within the scene.
```
marianogoldman, line 28:
```suggestion
- A) The entities MUST be unequivocal, and therefore their numbers MUST be unique identifiers within the scene.
```

content/ADR-166-entities-and-crdt-modification.md
```diff
@@ -0,0 +1,98 @@
...
+- B) The memory that defines the state of the scene MUST be a function of the number of existing entities at that moment and MUST NOT the number of entities created so far.
```
marianogoldman, line 29:
```suggestion
- B) The memory that defines the state of the scene MUST be a function of the number of existing entities at that moment and MUST NOT be the number of entities created so far.
```

content/ADR-166-entities-and-crdt-modification.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Solution space exploration
+
+In order to provide the versatility of destroying and creating entities, the `version` number is introduced, in this way each time an entity is destroyed the number is released and `version` is incremented by one. Finally, to make the entity unambiguous, the entity is defined as its tuple `<number, version>` regardless of how they are composed.
+
+So, up to here:
+1. The entity is no longer defined as a number, but it can continue to be defined as a number made up of the tuple `<number, version>` that composes it and thus be able to satisfy A).
```
marianogoldman, line 36:
```suggestion
1. The entity is no longer defined as a number, but as the tuple `<number, version>` that composes it and thus be able to satisfy A).
```

content/ADR-166-entities-and-crdt-modification.md
```diff
@@ -0,0 +1,98 @@
...
+    return { n , v }
+}
+
+function deleteEntity(entity: Entity) {
+    const { n , v } = entity
+    freeEntities.push(n)
+} 
+```
+
+If we create 1,000 entities and then remove and recreate as many times as you want the same amount, always staying below 1,000 , `counter will be equal to 1000`.
```
marianogoldman, line 66:
```suggestion
If we create 1,000 entities and then remove and recreate as many times as we want the same amount, always staying below 1,000, `counter will be equal to 1000`.
```

marianogoldman (changes requested):

PR #170, opened 2023-01-10
feat: ADR-170 subgraph worker

fzavalia:
ADR-170 for the subgraph cloudflare worker

### DIFF
--- a/content/ADR-170-subgraph-cloudflare-worker.md
+++ b/content/ADR-170-subgraph-cloudflare-worker.md
@@ -12,7 +12,11 @@ authors:
 
 ## Abstract
 
-Using a [Cloudfare Worker](https://workers.cloudflare.com/) to make graphql queries to subgraphs hosted by different providers to increase reliability.
+The Decentraland ecosystem relies on making graphql queries to different subgraphs to obtain indexed data from the blockchain. However, the use of a single provider for these subgraphs, TheGraph's Hosted Service, has led to issues such as subgraphs becoming corrupt or lagging behind, affecting the overall ecosystem. To address these issues and prepare for the potential sunset of TheGraph's service, a new solution was proposed: the use of a Cloudflare Worker as a provider proxy.
+
+By using a worker, different strategies can be coded to pick a provider and handle errors accordingly, abstracting this logic from the application level into a single endpoint. This allows for greater reliability by using a random provider and fallback options if the initial provider fails.
+
+All Decentraland applications should begin querying subgraphs through the worker, with the API remaining the same for seamless integration.
 
 ## Context, Reach & Prioritization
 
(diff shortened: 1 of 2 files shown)


content/ADR-170-subgraph-cloudflare-worker.md
```diff
@@ -0,0 +1,79 @@
...
+curl --request POST 
+  --url $WORKER_URL/collections-matic-mainnet \
+  --data '{"query":"{\n  nfts {\n    id\n  }\n}"}'
+```
+
+The configured strategy, called `RandomAndFallback` will forward the query to a random provider. If that random provider fails to be fetched or returns a non 200 response, the rest of the providers will be used as fallback until one responds correctly.
+
+## Solution Space Exploration
+
+All Decentraland applications should start querying subgraphs trough the worker. Meaning that in order to use it, they should update the current endpoints they are using to the new one. The API is exactly the same, so queries remain the same.
```
marianogoldman, line 51:
```suggestion
All Decentraland applications should start querying subgraphs through the worker. Meaning that in order to use it, they should update the current endpoints they are using to the new one. The API is exactly the same, so queries remain the same.
```

content/ADR-170-subgraph-cloudflare-worker.md
```diff
@@ -0,0 +1,79 @@
...
+status: Living
+type: Standards Track
+spdx-license: CC0-1.0
+authors:
+  - fzavalia
+---
+
+## Abstract
+
+Using a [Cloudfare Worker](https://workers.cloudflare.com/) to make graphql queries to subgraphs hosted by different providers to increase reliability.
```
marianogoldman, line 15:
```suggestion
The Decentraland ecosystem relies on making graphql queries to different subgraphs to obtain indexed data from the blockchain. However, the use of a single provider for these subgraphs, TheGraph's Hosted Service, has led to issues such as subgraphs becoming corrupt or lagging behind, affecting the overall ecosystem. To address these issues and prepare for the potential sunset of TheGraph's service, a new solution was proposed: the use of a Cloudflare Worker as a provider proxy.

By using a worker, different strategies can be coded to pick a provider and handle errors accordingly, abstracting this logic from the application level into a single endpoint. This allows for greater reliability by using a random provider and fallback options if the initial provider fails.

All Decentraland applications should begin querying subgraphs through the worker, with the API remaining the same for seamless integration.
```

marianogoldman: Totally optional, but I like that the abstract describes a bit of the problem, the proposed solution and the conclusions (so as to avoid having to read it all to pick up the big picture).

marianogoldman (approved):

PR #189, opened 2023-02-08
Social service M2 

jmoguilevsky:
Social service M2 implies lots of changes in the internal architecture of the service itself, Browser-interface, and Unity renderer. This ADR has the explanation of those changes.


content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,131 @@
...
+
+```mermaid
+sequenceDiagram
+    participant Renderer
+    participant Social client
+    participant Social service
+    participant Browser interface
+
+    Renderer ->>+ Social client: Get friends
+    Social client ->>+ Social service: Get friends
```
moliva, line 125: Make explicit this is the _new_ Social Client in C# and not the current one in TS.

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,131 @@
...
+    User 1 ->>+ Social service: Login
+    Social service ->>+ Matrix: Login
+    Matrix -->>- Social service: bearer token
+    Social service -->>- User 1: bearer token
+    User 1 ->>+ Social service: Connect (token)
+    Social service -->> User 1: Friends
+    Social service -->> User 1: Friend Requests
+    loop  Connection messages examples
+    User 1 ->> Social service: Friend event (ACCEPT, DELETE, etc.)
+    User 1 ->> Social service: Update position
```
moliva, line 85: Why do we need to add the update position here? This seems to still be processed in Matrix FTBB.

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,131 @@
...
+sequenceDiagram
+    participant User 1
+    participant Social service
+    participant Matrix
+
+    User 1 ->>+ Social service: Login
+    Social service ->>+ Matrix: Login
+    Matrix -->>- Social service: bearer token
+    Social service -->>- User 1: bearer token
+    User 1 ->>+ Social service: Connect (token)
```
moliva, line 80: Could we also clarify here that from this point forward all communications are through WebSocket and RPC?

agusaldasoro (approved):

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,165 @@
...
+    Social service ->> Matrix: create room
+    Social service -->>+ User 1: New FriendshipId: roomId
+    Social service -->>+ User 2: New FriendshipId: roomId
+```
+
+This new flow implies that rooms in Matrix will not be created until the friendship is established. This does not make the opposite true, if the friendship ends the room won't be destroyed (to prevent the history from being lost).
+
+Now, when a user logs in to decentraland, the friendship requests will also be obtained from the social service instead of the matrix sdk.
+
+First, the client will need to log in via HTTP, then establish a connection with the server, once correctly established, the server will then send two messages one with the current friends and another with the pending friend requests.
```
Julieta11, line 68:
```suggestion
First, the client will need to log in via HTTP. Then, establish a connection with the server. Once correctly established, the server will then send two messages: one with the current friends and another with the pending friend requests.
```

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,165 @@
...
+    end
+    User 1 ->> Social service: Disconnect WebSocket
+    Social service ->>- User 1: Disconnect WebSocket
+```
+
+## Implementation
+
+### Unity Renderer changes
+
+Given that Browser-interface is going to be slowly migrated into unity's codebase, the implementation of this feature will be done directly on unity's side too. Since this feature includes friends and friend requests, all of the code needed to handle friends and friendship requests will also be moved to unity, decoupling the dependency on Browser-interface. Still, channels, presence, and messaging will stay on the Browser interface side, but within time they will also be migrated to unity and will be able to leverage the new communication system.
```
Julieta11, line 95:
```suggestion
Given that Browser-interface is going to be slowly migrated into Unity's codebase, the implementation of this feature will also be done directly on Unity's side. To reduce the dependency on Browser-interface, the code responsible for handling friends and friend requests will be moved to Unity as part of this feature's implementation. However, channels, presence, and messaging will remain on the Browser-interface side for now. Over time, these features will also be migrated to Unity and will be able to take advantage of the new communication system
```

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,165 @@
...
+
+To prevent migrating all the code into unity-renderer's code and adding a lot of logic, this ADR proposes to add a new library in C# that will handle the logic around friendships and modify the logic in unity renderer to become more of an orchestrator of events.
```
Julieta11, line 97:
```suggestion
To prevent migrating all of the code into Unity Renderer's code and adding a lot of logic, this ADR proposes adding a new C# library to handle the logic around friendships. The logic in Unity Renderer would be adjusted to focus on event orchestration rather than handling all of the logic directly.
```

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,165 @@
...
+    Browser interface -->>- Renderer: Profiles[]
+    Renderer ->> Renderer: Store profiles in catalog
+```
+
+In this example, there are two main responsibilities of the flow:
+
+1. Obtaining the friendships list
+1. Storing the profiles in the catalog (to be able to render those friends)
+
+It can be seen that in the first example it's the browser interface's responsibility to know what profiles are missing and then send them to unity, making the browser interface the information orchestrator, in this new approach, Renderer is the orchestrator.
```
Julieta11, line 136:
```suggestion
In the first example, it is the Browser Interface's responsibility to identify which profiles are missing and then send this information to Unity. This makes the Browser Interface the information orchestrator. In the new approach, Unity Renderer would take on this role of information orchestration. 
```

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Browser interface changes
+
+This new structure will also imply changes on the browser interface side. First, all of the friendship logic will be deleted, maintaining only the necessary state so that the channels and messaging code can still function without any bugs. What will be deleted are the interactions with Unity around storing the friendships, friendship requests, and the logic for fetching profile information before sending the friends.
```
Julieta11, line 140:
```suggestion
This new structure will also require changes on the Browser Interface side. First, all of the friendship logic will be deleted, except for the necessary state to ensure that channels and messaging code continue to function without any issues. This will involve deleting the interactions with Unity around storing the friendships, friendship requests, and the logic for fetching profile information before sending the friends.
```

Julieta11 (commented):
Great work compis! I have some nitpicker suggestions which I think they might help with the readability :)

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,165 @@
...
+| Rust-RPC can’t handle the expected load                                            | Low                                             | High     | Invest time in developing the framework to make it more performant                                                                                                            | If this problem gets to prod, users might be impacted by not receiving all expected messages. Or in the worst case, take down the social service altogether | Medium        | 1        |
+| Rust RPC has internal bugs                                                         | Medium                                          | High     | Invest time in developing the framework to reduce bugs                                                                                                                        | If this problem gets to prod, users might be impacted by not receiving all expected messages. Or in the worst case, take down the social service altogether | High          | 1        |
```
marianogoldman, line 163: Choose either `Rust RPC` or `Rust-RPC` and use it consistent in all occurrences in this doc.

marianogoldman (changes requested):

content/ADR-189-social-service-M2.md
```diff
@@ -0,0 +1,166 @@
...
+Also, this new approach will need to be able to notify the users in real-time about the new friendship events. A new communication system will be needed to established to connect the client and the server (the technical specification of this system will be explored later).
+
+## Solution Space Exploration
+
+There are two main approaches:
+
+### WebSocket / Long polling
+
+WebSocket inherently comes with the benefit of establishing a bidirectional connection, enabling more features for the future (i.e. presence). Another big plus of using WebSocket is that the client can leverage protobuf and RPC for communication between the server and the client.
+On the other hand, WebSocket + RPC implies that the Social server will need to be the first implementation of Rust-RPC which is not yet a production-tested product.
```
lauti7, line 38: Why do you use"Rust-RPC" instead of "rpc-rust" (repo name) or "dcl-rpc" (crate name), or also a link to the project?

marianogoldman (approved):

PR #194, opened 2023-03-13
feat: quests PoC - user tracking RFC

guidota:


content/ADR-X-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,95 @@
...
+The `QuestsService` service definition includes three methods that allow users to start and abort quests and send events to the quests system. The `Subscribe` method returns a stream of `UserUpdate` messages that can be used to keep the client-side quests system up to date with the state of the user's quests and event processing.
+
+Once this client is implemented, it would be used to develop a Global Scene in charge of the Quest UI and user tracking logic. Using the SDK UI components is recommended, so it works on any Explorer implementation.
+
+### Quests UI
+For this PoC, the UI will show the list of current quests and their states, i.e., showing the completed and next steps for each quest with their respective descriptions. 
+
+### Quest User Tracking
+Ideally, users should not send repeated or non interesting events to the backend, to avoid extra communication and event processing. This piece of code would contain the logic to know which events are interesting and not yet sent, according to the current state of the user's quests. Finally, sending those events should be part of this scene code.
```
jmoguilevsky, line 90: Can we define what a non-interesting event is? 
On the repeated part of the events, say the update of the player's position is ping-ponging between two positions swiftly, should a client send every position change?

content/ADR-X-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,95 @@
...
+  string user_address = 1;
+}
+
+message StartQuestRequest {
+  string user_address = 1;
+  string quest_id = 2;
+}
+
+message StartQuestResponse {
+  bool accepted = 1;
```
jmoguilevsky, line 36: when would a start quest not be accepted? Can a quest be restarted?

content/ADR-X-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,95 @@
...
+  rpc Subscribe(UserAddress) returns (stream UserUpdate) {}
+}
+```
+
+The protobuf definition defines the messages and service necessary for the quests system:
+The `QuestsService` service definition includes three methods that allow users to start and abort quests and send events to the quests system. The `Subscribe` method returns a stream of `UserUpdate` messages that can be used to keep the client-side quests system up to date with the state of the user's quests and event processing.
+
+Once this client is implemented, it would be used to develop a Global Scene in charge of the Quest UI and user tracking logic. Using the SDK UI components is recommended, so it works on any Explorer implementation.
+
+### Quests UI
```
jmoguilevsky, line 86: Can you also add a link/description of the server implementation?

content/ADR-X-quests-poc-user-tracking.md
guidota: I'll clarify the non-interesting event. The idea is to not send events that may not make progress in any quests, so in the example of the positions, if any position is relevant to a current quest, it should only send it once, otherwise never.

content/ADR-X-quests-poc-user-tracking.md
guidota: There are a few valid reasons: 
- Quest is not found
- Quest is deactivated (the owner deleted it)
- User already started the quest
- Internal errors (DB connection failed or something like that)

A quest can't be restarted at the moment (I think aborting and starting it again would work since we don't validate that yet)

content/ADR-X-quests-poc-user-tracking.md
guidota: I'd rather not add external links that may be broken or get obsolete in the future. I was thinking that the idea of having a proto service definition is that the client is agnostic to server implementation, what do you say? (Also this document is focused on the client side only)

content/ADR-X-quests-poc-user-tracking.md
jmoguilevsky: I'd add this rules somewhere in the ADR

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Solution Space Exploration
+
+The client will use the Decentraland RPC and protocol library to communicate with the QuestsService backend. (WebSockets + ProtoBuffers) 
+
+```proto
+syntax = "3";
+package decentraland.quests;
+
+message UserAddress {
```
agusaldasoro, line 28: would it make sense to name it just `User` in case there's a need for more fields in the future?

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+  string user_address = 1;
+}
+
+message StartQuestRequest {
+  string user_address = 1;
+  string quest_id = 2;
+}
+
+message StartQuestResponse {
+  /* There are a few valid reasons to not be accepted:
```
agusaldasoro, line 38: should those reasons be in a message?

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+  *  - Quest is not found
+  *  - Quest is deactivated (the owner deleted it)
+  *  - User already started the quest
+  *  - Internal errors (DB connection failed or something like that) */
+  bool accepted = 1;
+}
+
+message AbortQuestRequest {
+  string user_address = 1;
```
agusaldasoro, line 47: is the user_address part of the payload? Shouldn't the user be already authenticated?

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+  /* There are a few valid reasons to not be accepted:
+  *  - Quest instance is not found
+  *  - Quest instance is from another user 
+  *  - Quest instance already aborted
+  *  - Internal errors (DB connection failed or something like that) */
+  bool accepted = 1;
+}
+
+message Event {
+  string address = 1;
```
agusaldasoro, line 61: should this be `User`/`UserAddress`?

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+  repeated Action action_items = 3;
+}
+
+message StepContent {
+  repeated Task to_dos = 1;
+  repeated string task_completed = 2;
+}
+
+message QuestState {
+  map<string, StepContent> current_steps = 1;
```
agusaldasoro, line 88: what does represent the key of this map?

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+}
+
+message EventResponse {
+  option fixed32 event_id = 1;
+  bool accepted = 2;
+}
+
+message Action {
+  string type = 1;
+  map<string, string> parameters = 2;
```
agusaldasoro, line 73: could you comment with an example of key+value here?

content/ADR-194-quests-poc-user-tracking.md
lauti7, line 38: typo: `not to be accepted`

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+## Context, Reach & Prioritization
+
+The implementation of a quests system in a virtual world or game can greatly enhance the player experience and provide a sense of achievement and progress. This feature is particularly important for virtual worlds and games that have an open-world structure and rely on player-generated content. A client-side quests system can help guide players towards engaging with the available content and provide incentives for players to create their own content.
+
+## Solution Space Exploration
+
+The client will use the Decentraland RPC and protocol library to communicate with the QuestsService backend. (WebSockets + ProtoBuffers) 
+
+```proto
+syntax = "3";
```
lauti7, line 25: this should be `syntax = "proto3"`

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+  oneof message {
+    repeated QuestState quest_state = 1,
+    fixed32 event_ignored = 2;
+  }
+}
+
+service QuestsService {
+  // User actions
+  rpc StartQuest(StartQuestRequest) returns (StartQuestResponse) {}
+  rpc AbortQuest(AbortQuestRequest) returns (AbortQuestResposne) {}
```
lauti7, line 104: typo: `AbortQuestResposne ` -> `AbortQuestResponse`

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+}
+
+message Event {
+  string address = 1;
+  string type = 2; 
+  map<string, string> parameters = 3;
+}
+
+message EventResponse {
+  option fixed32 event_id = 1;
```
lauti7, line 67: it's `optional`

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,132 @@
...
+message QuestState {
+  map<string, StepContent> current_steps = 1;
+  fixed32 steps_left = 2;
+  repeated string steps_completed = 3;
+  repeated string required_steps = 4;
+}
+
+message UserUpdate {
+  oneof message {
+    repeated QuestState quest_state = 1,
```
lauti7, line 96: replace comma at the end with semicolon

agusaldasoro (approved):

content/ADR-194-quests-poc-user-tracking.md
```diff
@@ -0,0 +1,141 @@
...
+}
+
+service QuestsService {
+  // User actions
+  rpc StartQuest(StartQuestRequest) returns (StartQuestResponse) {}
+  rpc AbortQuest(AbortQuestRequest) returns (AbortQuestResponse) {}
+  rpc SendEvent(Event) returns (EventResponse) {}
+
+  // Listen to changes in quest states and event processing updates
+  rpc Subscribe(UserAddress) returns (stream UserUpdate) {}
```
pbosio, line 119: shouldn't that `UserAddress` be `User`?
is `User` message only intended for the `Subscribe` service method?
because, if that's the case, maybe renaming it for something like `SubscribeRequest` might avoid any confusion when having all those messages requiring an `user_address`

menduz (approved):

content/ADR-194-quests-poc-user-tracking.md
guidota: Yes! This definition would probably change a few more times on early iterations, so it's just the estimated interaction

PR #197, opened 2023-03-14
transform component updates

menduz:


content/ADR-153-transform-sdk-component.md
```diff
@@ -0,0 +1,121 @@
...
+parameters:
+  COMPONENT_ID: 1
+  COMPONENT_NAME: core::Transform
+  CRDT_TYPE: LastWriteWin-Element-Set
+```
+
+The Transform component is serialized using a plain-old C struct due to the amount of times it needs to be serialized and deserialized, it uses a fixed memory layout, represented as `TransformComponent` in the following snippet:
+
+```c++
+struct vec3 {
```
marianogoldman, line 33: Shouldn't this be `Vec3`... trying to be consistent with `TransformComponent`. Same for `cuaternion`

content/ADR-153-transform-sdk-component.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+All fields are encoded in little-endian.
+
+## Semantics of parenting
+
+The parenting of entities is a complex problem because in an ECS approach, entities are stored in a flat structure, and trees are a synthetic construct for positioning purposes only. The decision is that when an entity in the middle of a branch is removed, all its children will be re-parented to the root entity.
+
+One of the complexities is that rendering engines usually require a tree-like structure to calculate all the world matrixes. Due to the commutative nature of the CRDT messages, there may be scenarios in which during a window of time the reflected state of the messages contain cycles.
```
marianogoldman, line 53:
```suggestion
One of the complexities is that rendering engines usually require a tree-like structure to calculate all the world matrices. Due to the commutative nature of the CRDT messages, there may be scenarios in which, during a window of time, the reflected state of the messages contain cycles.
```

content/ADR-153-transform-sdk-component.md
```diff
@@ -0,0 +1,121 @@
...
+
+The suggested implementation path is to move the entities to the root level of the scene while there are parenting cycles. Always prioritizing the best possible performance for the best case scenario (state without cycles). If the scenes are well programmed, after processing all CRDT messages the scene should converge towards a cycle-less DAG starting on the root entity.
+
+To ellaborate on the parenting process, we must first introduce how vertex projection works on 3D engines. It is all based on matrix calculations. Starting from an identity matrix, we can translate, scale or rotate the matrix by multiplying the same matrix by a rotated identity or rotated scale matrix, as many times as needed.
```
marianogoldman, line 57:
```suggestion
To elaborate on the parenting process, we must first introduce how vertex projection works on 3D engines. It is all based on matrix calculations. Starting from an identity matrix, we can translate, scale or rotate the matrix by multiplying the same matrix by a rotated identity or rotated scale matrix, as many times as needed.
```

content/ADR-153-transform-sdk-component.md
```diff
@@ -0,0 +1,121 @@
...
+
+- When a `Transform` component doesn't exist in an entity, it is assumed that the entity has a `scale={1,1,1} and rotation={0,0,0,1}`.
+- Thus, it is safe to assume that a missing `Transform` component has the same rendering semantics than an entity with a `Transform.Identity`.
+- And since entities are defined by the set of components and don't exist by themselves, it is safe to assume that all entities have a `Transform.Identity` by default.
+- Thus we can deduce that parenting an entity A with a parent B (which doesn't exist, because it doesn't have any components) should have the same semantics as parenting an entity A with a valid parent entity B with a `Transform.Identity` component.
+
+#### When an entity A is added to an entity B but the engine does not know about entity B
+
+- The engine MUST assume that the entity B is valid, even though "it doesn't exist" because no component is registered to it.
+- The assumption for the `Transform` component of the entity B is that the entity have a `Transform.Identity` component.
```
marianogoldman, line 86:
```suggestion
- The assumption for the `Transform` component of the entity B is that the entity has a `Transform.Identity` component.
```

marianogoldman (approved):

content/ADR-153-transform-sdk-component.md
menduz: done

PR #200, opened 2023-03-14
raycast component

menduz:


content/ADR-200-raycast-sdk-component.md
```diff
@@ -0,0 +1,209 @@
...
+
+That implies that raycast MUST be executed after the "world matrix" of the entity in which they exist is calculated and not dirty. This is of extreme importance, because the ray takes its origin point based on the _worldMatrix_ of the entity. Many interesting properties are inherited from this design, like raycasts that react to billboard rotations or raycasts that are attached to bones and animations of the avatars.
+
+The origin point and global direction are calculated as follow:
+
+```typescript
+// first calculate the global origin of the raycast
+const globalOrigin = entity.worldMatrix.multiply(vec4(raycast.originOffset.xyz, 1))
+
+// and then calculate the global direction, relative to the
```
LautaroPetaccio, line 139: It seems that the phrase is incomplete, would you mind completing it?

content/ADR-200-raycast-sdk-component.md
```diff
@@ -0,0 +1,209 @@
...
+}
+```
+
+### Usage of the `timestamp` property
+
+The timestamp property is a correlation number, only defined by the scene. The renderer MUST copy the value of the `timestamp` from the Raycast component to the RaycastResult component.
+
+### Performance considerations
+
+Having multiple continuous raycasts in a scene may have a big performance penalty. As a scene developer, it is RECOMMENDED to keep this number at minimum. As a renderer implementator, the compute of raycasts MUST count towards the execution quota of each scene to prevent a non-optimized scene from affecting the overall quality of the experience.
```
LautaroPetaccio, line 194: What if we move this section out of semantics and move it next to the RFC sections as a `##` header?

PR #203, opened 2023-03-31
Create websocket-login.md

agusaldasoro:
ADR for discussions regarding authentication in Social Service Websocket.


content/ADR-203-social-ws-login.md
```diff
@@ -0,0 +1,158 @@
...
+- kuruk-mm
+
+---
+
+## Abstract
+
+This document contains the analysis of different solutions for authenticating a client and managing the token required for accessing the FriendshipsService. Each solution will be evaluated based on its pros and cons, including aspects such as scalability, security, and complexity.
+
+This is an exhaustive analysis for the Authentication section of [Social Service M2](/adr/ADR-189).
+
```
agusaldasoro, line 26:
```suggestion
## Abstract

This document contains the analysis of different solutions for authenticating a client and managing the token required for accessing the FriendshipsService. Each solution will be evaluated based on its pros and cons, including aspects such as scalability, security, and complexity.

This is an exhaustive analysis of the Authentication section of [Social Service M2](/adr/ADR-189).

```

content/ADR-203-social-ws-login.md
```diff
@@ -0,0 +1,158 @@
...
+- If a new connection with the same token/user_id arrives, the previous connection can be closed. This is a simple way to avoid DoS.
+
+#### Solution 1: Disadvantages
+
+- The information is stored in memory, which may be a replication of what is in Redis.
+- [dcl-rpc](https://github.com/decentraland/rpc-rust) needs to be expanded to allow connection identification and specific information storage. Currently, a global context is stored for the entire service, but a way to identify the client (there is no IP address or socket, only user_id) needs to be found. Additionally, `dcl-rpc` currently does not expose the connection because it is abstracted.
+- Client-side logic is more complex, as it must ensure that the first message sent is the login message.
+- As the service has a state, then it makes more complex to scale horizontally, as a new login is required when reconnecting with another node, even using the same token.
+- Websockets support automatic reconnection, in this case if the server restarts and loses the memory state and reconnects to the user, then the login information will be lost.
```
agusaldasoro, line 62:
```suggestion
- The information is stored in memory, which may be a replication of what is in Redis.
- [dcl-rpc](https://github.com/decentraland/rpc-rust) needs to be expanded to allow connection identification and specific information storage. Currently, a global context is stored for the entire service, but a way to identify the client (there is no IP address or socket, only user_id) needs to be found. Additionally, `dcl-rpc` currently does not expose the connection because it is abstracted.
- Client-side logic is more complex, as it must ensure that the first message sent is the login message.
- As the service has a state, then it makes it more complex to scale horizontally, as a new login is required when reconnecting with another node, even using the same token.
- Websockets support automatic reconnection, in this case, if the server restarts and loses the memory state and reconnects to the user, then the login information will be lost.
```

content/ADR-203-social-ws-login.md
```diff
@@ -0,0 +1,156 @@
...
+
+#### Solution 4: Advantages
+
+- Sends the request to Redis/Synapse only once
+- Does not allow more than one connection per user (prevents simple DOS)
+- No need to send the token as part of the message
+
+#### Solution 4: Disadvantages
+
+- The client must support cookies or be able to send the authorization header.
```
menduz, line 152: There is currently so much struggle while mixing cookies and websockets. It is not a straight forward implementation in _most_ languages.

Besides the ergonomics of getting cookies to work, I think setting cookies is something very web 1.0 for native public-key applications like Decentraland. I'd like to propose the ws-comms authentication mechanism for the WS connection. It is implemented in `bff`, `ws-room-service` and `mini-comms` repositories. 

```mermaid
sequenceDiagram
  client-->>server: open ws
  server-->>client: pleaseSignChallenge(challengeText=randomString())
  client-->>server: signedChallenge(result=sign(challengeText))
  server-->>server: authenticate, authorize
  server-->>server: load modules based on the permissions
  server-->>client: welcomeMessages(avalableModulesList)
```

The messages are simple and straight forward to implement, and the solution without cookies adds an extra layer of security against session spoofing

[BFF proto definitions](https://github.com/decentraland/protocol/blob/main/proto/decentraland/bff/authentication_service.proto#L24-L27)

```protobuf
syntax = "proto3";
package decentraland.bff;

message GetChallengeRequest {
  string address = 1;
}

message GetChallengeResponse {
  string challenge_to_sign = 1;
  bool already_connected = 2;
}

message SignedChallenge {
  string auth_chain_json = 1;
}

message WelcomePeerInformation {
  string peer_id = 1;

  // list of available modules in this BFF
  repeated string available_modules = 2;
}

service BffAuthenticationService {
  rpc GetChallenge(GetChallengeRequest) returns (GetChallengeResponse) {}
  rpc Authenticate(SignedChallenge) returns (WelcomePeerInformation) {}
}
```

The tests are also super straight forward https://github.com/decentraland/explorer-bff/blob/4352737ca24a0c590387fba4b46da297408e899a/test/integration/rpc.spec.ts#L18-L31

This has a clear advantage of not needing to store a second token, only making use of the already available auth chain to authenticate the users. It will also enable interoperability with the [scoped permissions for auth chains](https://github.com/decentraland/adr/pull/162/files)

menduz (changes requested):
Continuing with the comment of the PR, the more I think of it the more it makes sense to use some AuthenticationRpc with signed fetch like the bff does. Here are some reasons:
- [AuthChain](https://docs.decentraland.org/contributor/auth/authchain/) signatures is the unequivocal way to authenticate user in Decentraland, the `/login` endpoint would have to use the AuthChain anyways. Diverging from the canonical mechanism wouldn't be sensible on this scenario.
- Authenticating the transport (ws via cookie) would not enable change of transports. Even though the case of changing WS is unlikely going to happen, it is the same as TLS vs Application layer security
- Native implementations of WS do not handle cookies correctly, using an RPC would make it way easier to implement other clients.
- So far we have never used cookies or long lived token because of the security implications of what stealing the tokens can cause. 


At the end of the day, I think implementing the authentication using RPC will reduce steps in the implementation and improve security. This solution also plays very well with the original vision of the ADR-143, keeping matrix tokens accessible to the server only.

content/ADR-203-social-ws-login.md
```diff
@@ -0,0 +1,287 @@
...
+- `2.b: the login message is sent by the Server`
+- `3: The Client logins against the service`
+
+Among those solutions, the benefit of implementing the third one is that the Synapse Token is opaque to the client, making it something internal that can be removed in the future if Synapse dependency is removed. Anyway, the cost of implementing that solution today implies many changes in new dcl libraries like `rpc-rust` and `decentraland-crypto-rust` which may imply a risk for a project as big as Social Service M2 is.
+
+For the Scope of the Social Service M2 Project, the decition is to implement the `2.c` solution, which will handle the authentication individually on each message, as it is the cheaper the implement and the focus and effort is in the robustness and availability of a new Websocket Server implemented in Rust that is thought to replace the dependency to Synapse when managing friendships and request friendships. The `2.c` solution does not represent any security risk or concern, as the Synapse Token is already stored on the local storage of the client, and as it will be validated on every message the behaviour will be analogue on what exists today.
```
agusaldasoro, line 285:
```suggestion
For the Scope of the Social Service M2 Project, the decision is to implement the `2.c` solution, which will handle the authentication individually on each message, as it is cheaper the implement and the focus and effort is in the robustness and availability of a new WebSocket Server implemented in Rust that is thought to replace the dependency to Synapse when managing friendships and request friendships. The `2.c` solution does not represent any security risk or concern, as the Synapse Token is already stored on the local storage of the client, and as it will be validated on every message the behavior will be analog to what exists today.
```

content/ADR-203-social-ws-login.md
```diff
@@ -0,0 +1,287 @@
...
+
+In this line, the idea is to revist this solution after Social Service M2 is released and then decide if a better Authentication management can be done (like implementing the `3` solution). So, the `.proto` definition used to manage the friendships in the Social Service for M2 is for internal use only, and the release will be marked as beta.
```
agusaldasoro, line 287:
```suggestion
In this line, the idea is to revise this solution after Social Service M2 is released and then decide if better Authentication management can be done (like implementing the `3` solution). So, the `.proto` definition used to manage the friendships in the Social Service for M2 is for internal use only, and the release will be marked as beta.
```

menduz (approved):
Amazing, although I think 3 is the way to go, it looks like a sensible plan. As stated on a comment in the conclusion, I wouldn't underestimate the cost of changing the authentication mid-flight. Please make sure to create messages with strictly optional authentication parts so it can be safely removed after.

content/ADR-203-social-ws-login.md
```diff
@@ -0,0 +1,287 @@
+
+---
+layout: adr
+adr: 203
+title: Authentication Mechanism for Friendship WebSocket Service
+date: 2023-03-31
+status: Draft
```
Julieta11, line 7: Commenting to point out the status. I'm not sure which one is correct.

Julieta11 (approved):

PR #220, opened 2023-04-19
Updates ADR-145 regarding realm-picking algorithm v2

aleortega:
Updates the ADR-145 which contains the enhancements to be done over the realm-picking algorithm.


content/ADR-145-pick-realm-v2.md
```diff
@@ -9,43 +9,63 @@ spdx-license: CC0-1.0
...
 
-### Accepting Users Filter
+### Overloaded Catalyst
 
 _Description_:
 
-Publicly exposing the Catalyst Node's available resources is a security risk, so their estimation should be done in the Catalyst (specifically in the BFF) and not in the Kernel.
-
-This way, the BFF will leverage the usage of the field `accepting_users` in the `/about` endpoint.
+Exposing the available resources of Catalyst Nodes publicly poses a security risk, and thus, their estimation should be conducted within the Catalyst itself (_specifically in the BFF_), rather than in the Kernel.
```
marianogoldman, line 63: Should we still mention BFF here?

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,129 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+export type ForceCatalystConfig = {
+  config?: { sortedOptions?: [string]; }
 }
 ```
+
+### Large Latency
+
+_Description_:
+
+This rule will still operate as it currently does. However, in order to implement it on the server-side, the client will need to check the latency on all nodes to pass this information to the service. To prevent clients from constantly calling all nodes to check latency, they will first check if the realm-picking service is using a `variant` that contains the `LARGE_LATENCY` filter and only if it does it will gather the latency information to pass it over.
```
marianogoldman, line 160: Does this mean an extra request to the server?

content/ADR-145-pick-realm-v2.md
aleortega: Currently this is done in the BFF, may it is worth keeping it like this for now and update the ADR once we completely remove the BFF, what do you think?

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
 adr: 145
 title: Pick Realm Algorithm V2
 date: 2022-11-28
-status: Draft
+status: Living
```
menduz, line 6: We only make ADR living when they are running in production

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
 _Pseudo Code_:
 
 ```typescript
-export function versionLink() {
-  const catalystsWithVersion: { realm: string, version: string } = peers.sortByVersion()
-  const versionBuckets: PriorityQueue<{ realm: string, version: string }[]> = groupByVersions(catalystsWithVersion)
-  // This selects randomly between the candidates in the selected bucket 
-  // And chooses the bucket with the same algorithm used for close peers
-  return selectFirstByScore(context, score, definitiveDecisionThreshold)
+export function catalystVersionLink(configuration: ForceCatalystConfig) {
```
menduz, line 134: If possible I'd try to use a different analogy than the current algorithm "links" it has proven hard to reason and understand

content/ADR-145-pick-realm-v2.md
agusaldasoro: If the BFF is about to be removed, then we shouldn't mention it here. Anyways, we should specify here who will be responsable for this

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
 
-This way, the BFF will leverage the usage of the field `accepting_users` in the `/about` endpoint.
+Consequently, the BFF will utilize the `acceptingUsers` field in the `/about` endpoint to leverage the calculation of this value. The logic on the Kernel side will be straightforward, involving filtering out Catalysts with a value of false for this field.
 
-The logic from the Kernel side will be simple: it will filter out the catalysts having that value as false.
-
-On the other hand, the BFF will have a set of rules to calculate that value. The `accepting_users` field will be false if:
-- The max amount of users for the server is reached
+On the other hand, the BFF will implement a set of rules to determine the value of `acceptingUsers`. The `acceptingUsers` field will be set to false if:
+- The maximum number of users for the server has been reached, with the maximum amount being specified through environment variables.
```
agusaldasoro, line 68: Should we limit the number of users before reaching the maximum? For example allowing a 90% of capacity

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
 - The CPU is at 90% or higher
 - The memory usage is at 90% or higher
```
agusaldasoro, line 71: The pseudo-code below is no longer valid, as that implementation lives in Kernel. Please remove it

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
   return undefined
 }
 ```
 
-_Configuration_:
+### Catalyst Version
+
+_Description_:
+
+Every Catalyst node provides information about the versions of its Content, Lambdas, BFF, and Comms components. The Kernel reads these version fields and compares them with the minimum required version specified in the rule configuration. As a result, the configuration has the flexibility to specify minimum required versions for one or more services, allowing the Catalyst to be considered as a candidate only if it meets these requirements. This means that the configuration file can specify minimum required versions for X and Xn services, giving the flexibility to specify versions for one or more services as needed.
```
agusaldasoro, line 94: Same here, it continues referencing to Kernel

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+  Object.keys(minimumRequiredVersions).forEach(([serviceName, serviceRequiredVersion]) => {
+    validCandidates = picked.filter(candidate => meetsVersion(candidate, serviceName, serviceRequiredVersion))
+    // ...
+  })
+
+  return validCandidates
 }
 ```
 
+_Configuration_:
```
agusaldasoro, line 112: I'm not sure if it's clear where will be this configuration stored and who will be in charge of it

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+  config?: 
+          { comms?: string;
+            bff?: string;
+            content?: string;
+            lambdas?: string;
+          }
+}
+```
 
+### Force Catalyst
```
agusaldasoro, line 125: The force catalyst will return only one? Does it make sense to return a recommended order instead?

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+export type ForceCatalystConfig = {
+  config?: { sortedOptions?: [string]; }
+}
+```
+
+### Large Latency
+
+_Description_:
+
+This rule will still operate as it currently does. However, in order to implement it on the server-side, the client will need to check the latency on all nodes to pass this information to the service. To prevent clients from constantly calling all nodes to check latency, they will first check if the realm-picking service is using a `variant` that contains the `LARGE_LATENCY` filter and only if it does it will gather the latency information to pass it over.
```
agusaldasoro, line 160: I'd change the responsibility here, instead of the client sending to the server their own latencies for each option, I'd prefer something like the server returning a recommended ordered list of catalysts and then the client will choose from that list according to its local latency.

For example:

```
Client ->  Server  // getListOfCatalyst() -> [catalystB, catalystD, catalystA]
Then the client chooses the first one on the list with an adequate latency
```

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+- Filter by OVERLOADED_CATALYST
+  - Firstly, nodes that are struggling to perform as expected will be filtered out.
+- Filter by LARGE_LATENCY
+  - Next, nodes that are located far away from the user in terms of latency will be filtered out.
+- Score by CLOSE_PEERS_SCORE
+  - After filtering, if the user is attempting to connect to a specific scene, they will be redirected to the node with the highest number of peers in that scene.
+- Score by ALL_PEERS_SCORE
+  - If the previous score does not result in the selection of a single catalyst, the node with the highest number of peers connected to it will be chosen.
+- Balance with LOAD_BALANCING
+  - As a fallback, if multiple catalysts are eligible for selection, a Round-Robin strategy will be implemented to evenly distribute the load among them.
```
agusaldasoro, line 176: nit:

```suggestion
1. Filter by OVERLOADED_CATALYST
  - Firstly, nodes that are struggling to perform as expected will be filtered out.
2. Filter by LARGE_LATENCY
  - Next, nodes that are located far away from the user in terms of latency will be filtered out.
3. Score by CLOSE_PEERS_SCORE
  - After filtering, if the user is attempting to connect to a specific scene, they will be redirected to the node with the highest number of peers in that scene.
4. Score by ALL_PEERS_SCORE
  - If the previous score does not result in the selection of a single catalyst, the node with the highest number of peers connected to it will be chosen.
5. Balance with LOAD_BALANCING
  - As a fallback, if multiple catalysts are eligible for selection, a Round-Robin strategy will be implemented to evenly distribute the load among them.
```

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+
+This variant will be useful when we want to crowd the nodes so the users can interact between them more frequently.
+- Score by CLOSE_PEERS_SCORE
+  - After filtering, if the user is attempting to connect to a specific scene, they will be redirected to the node with the highest number of peers in that scene.
+- Score by ALL_PEERS_SCORE
+  - If the previous score does not result in the selection of a single catalyst, the node with the highest number of peers connected to it will be chosen.
+- Balance with LOAD_BALANCING
+  - As a fallback, if multiple catalysts are eligible for selection, a Round-Robin strategy will be implemented to evenly distribute the load among them.
+
+### Server-Side Implementation
```
agusaldasoro, line 205: What will happen if the server to choose realm is down? Can the client execute the same logic for itself? What will be the default behavior?

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+}
+```
+
+### Large Latency
+
+_Description_:
+
+This rule will still operate as it currently does. However, in order to implement it on the server-side, the client will need to check the latency on all nodes to pass this information to the service. To prevent clients from constantly calling all nodes to check latency, they will first check if the realm-picking service is using a `variant` that contains the `LARGE_LATENCY` filter and only if it does it will gather the latency information to pass it over.
+
+### Variants
```
agusaldasoro, line 162: Does it make sense for the client to choose the variant that wants to be executed? 
For example: should the lambda receive the configuration as payload/params? And return the best choice of realm according to that?

content/ADR-145-pick-realm-v2.md
```diff
@@ -59,46 +79,137 @@ https://github.com/decentraland/kernel/blob/main/packages/shared/dao/index.ts#L6
...
+- Score by CLOSE_PEERS_SCORE
+  - After filtering, if the user is attempting to connect to a specific scene, they will be redirected to the node with the highest number of peers in that scene.
+- Score by ALL_PEERS_SCORE
+  - If the previous score does not result in the selection of a single catalyst, the node with the highest number of peers connected to it will be chosen.
+- Balance with LOAD_BALANCING
+  - As a fallback, if multiple catalysts are eligible for selection, a Round-Robin strategy will be implemented to evenly distribute the load among them.
+
+### Server-Side Implementation
+
+In this section, we will analyze the options for implementing the algorithm on the server side, considering factors such as global availability, cache capabilities, scalability, and cold-start performance. The two options under consideration are `AWS Lambdas` and `CloudFlare Workers`.
```
agusaldasoro, line 207: How will the cache work? Are workers or lambdas able to do that? Based on what parameters will be the cache function on?

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
+- Addition of new rules:
+  - Overloaded Catalyst: Servers with CPU or memory usage exceeding 90%, or reaching maximum user capacity, will be excluded from selection.
+  - Catalyst Version: Nodes with a version lower than the minimum required version (_as specified in the configuration_) will be excluded from selection. This will allow us to set minimum required version for each service running within a Catalyst.
+  - Catalyst Force: Possibility to forcibly select a Catalyst.
+- Creation of multiple default rule sets (known as _variants_) that can be easily switched to use different techniques for realm picking.
+- Relocating the algorithm to the server-side to accommodate multiple client implementations without losing any capabilities.
 
+From now on, the rules will be referred to as `links` and `filters`. Filters will be used to remove nodes that do not meet specific criteria, while links will be used to assign scores to nodes to determine the best one for redirecting users.
 
 ## Solution Space Exploration
```
agusaldasoro, line 34: It's not clear to me where will be the configuration file live, also who will be in charge of it? is it the client's responsibility? Is it a configuration of the centralized server/lambda/worker?

agusaldasoro (changes requested):

content/ADR-145-pick-realm-v2.md
hugoArregui: I won't mention the BFF, we can mention the realm description without specifying who servers, specially since the endpoint is `/about` not `/bff/about`, so bff is just an implementation detail here

aleortega: Got it, I'm removing this mention about the BFF. Is it enough to mention that these versions are going to be retrieved from each Catalyst?

content/ADR-145-pick-realm-v2.md
aleortega: Got it, moving this back to Draft, thanks!

content/ADR-145-pick-realm-v2.md
aleortega: I'm replacing this link concept for rules.

content/ADR-145-pick-realm-v2.md
aleortega: Done here -> https://github.com/decentraland/adr/pull/220/commits/3d60c885159e2dd9a286a48b3865cdfa3fdce0cf

content/ADR-145-pick-realm-v2.md
aleortega: Done here -> [feedback: replace links concept by rules](https://github.com/decentraland/adr/pull/220/commits/8a31761d537647dc8168a52bbebb6572da603234)

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
 
-The logic on the realm scoring algorithm should be improved to take into consideration more variables in order for a server to be more eligible. As with the current model, any added variable needs to be modifiable through Feature Flags so that they can be changed in case of need without a release or a reboot of the server.
+With regard to functionality, the algorithm will be updated to incorporate additional variables in the Catalyst selection process through the addition of new rules. As for the implementation, the software will be relocated to the server-side to accommodate multiple client implementations.
 
 ## Context, Reach & Prioritization
 
-Some aspects that should be considered:
+The realm-picking algorithm utilizes a predefined set of rules called `variants` to dynamically switch to the optimal approach based on the platform context through configuration files. However, it has been identified that this algorithm lacks certain capabilities to address potential contexts that Decentraland may encounter. To enhance the selection algorithm, the following improvements are intended:
```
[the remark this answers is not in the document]
aleortega: Done here -> [feedback: update Context, Reach & Prioritization section](https://github.com/decentraland/adr/pull/220/commits/0640cc785771b031824a5a457aa939ad62f0566b)

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
 
-- Resources: if a server has either CPU or memory at 90%, don't add more load to that server.
-- Catalyst version: if the version of the node is behind the network, it should be less eligible.
+- Addition of new rules:
+  - Overloaded Catalyst: Servers with CPU or memory usage exceeding 90%, or reaching maximum user capacity, will be excluded from selection.
+  - Catalyst Version: Nodes with a version lower than the minimum required version (_as specified in the configuration_) will be excluded from selection. This will allow us to set minimum required version for each service running within a Catalyst.
+  - Catalyst Force: Possibility to forcibly select a Catalyst.
```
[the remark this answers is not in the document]
aleortega: Done here -> [feedback: update Context, Reach & Prioritization section](https://github.com/decentraland/adr/pull/220/commits/0640cc785771b031824a5a457aa939ad62f0566b)

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
+- Creation of multiple default rule sets (known as _variants_) that can be easily switched to use different techniques for realm picking.
```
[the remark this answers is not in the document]
aleortega: Done here -> [feedback: update Context, Reach & Prioritization section](https://github.com/decentraland/adr/pull/220/commits/0640cc785771b031824a5a457aa939ad62f0566b) and here [feedback: replace links concept by rules](https://github.com/decentraland/adr/pull/220/commits/8a31761d537647dc8168a52bbebb6572da603234)

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
-- Accepting Users Filter: Will filter out if the amount of resources for that Catalyst is at limit.
-- Version Link: the algorithm will prioritize sending traffic to instances that have the latest version
+### Algorithm improvement
 
+The existing implementation of the realm picking algorithm, as outlined in the [Realm Picking Algorithm ADR](https://adr.decentraland.org/adr/ADR-86), allows for extensibility through the addition of new links and filters. In this proposal, two filters and a new link will be introduced:
+- **OVERLOADED_CATALYST**: This filter will exclude a Catalyst if its resource usage or user capacity reaches a certain threshold.
+- **VERSION_CATALYST**: This filter will exclude a Catalyst if it does not meet the minimum required version as specified.
+- **FORCE_CATALYST**: This new link will enable the forcible selection of a Catalyst listed in its configuration.
+
+Furthermore, four new variants will be added, each containing a set of rules designed to address specific contexts.
```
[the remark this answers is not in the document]
aleortega: Done here -> [feedback: rephrase Algorithm improvement section](https://github.com/decentraland/adr/pull/220/commits/40bc236e7c9360c24ef6cf046ef6d45676583758)

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
+
+### Server-side
```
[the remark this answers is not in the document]
aleortega: Done here -> [feedback: rename Server-Side title for Allocating the algorithm on it…](https://github.com/decentraland/adr/pull/220/commits/b94f59fe348bf09b959e5595fdbfaa8db7b3f81b)

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
+
+This algorithm will be implemented on the server-side to accommodate various client implementations. While this may require modifying the implementation of the `LARGE_LATENCY` rule to receive necessary information from the client-side, it will enable enhancements such as adding a cache mechanism or event-driven notifications to proactively filter out unhealthy Catalysts, thereby improving the overall performance of the algorithm.
```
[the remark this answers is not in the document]
aleortega: Explained here -> [feedback: explain how LARGE_LATENCY rule will work and replace varian…](https://github.com/decentraland/adr/pull/220/commits/7a826fc7ae1a92be6f453db0aba3ef24b940ba3b)

Currently this rule filters out the nodes that do not meet a certain threshold regarding latency.

content/ADR-145-pick-realm-v2.md
```diff
@@ -3,49 +3,69 @@ layout: adr
...
+- Relocating the algorithm to the server-side to accommodate multiple client implementations without losing any capabilities.
 
+From now on, the rules will be referred to as `links` and `filters`. Filters will be used to remove nodes that do not meet specific criteria, while links will be used to assign scores to nodes to determine the best one for redirecting users.
 
 ## Solution Space Exploration
 
-The current implementation (as described in [Realm Picking Algorithm ADR](/adr/ADR-86)) is expansible by adding new links. In this proposal a filter and a new link will be added: 
-- Accepting Users Filter: Will filter out if the amount of resources for that Catalyst is at limit.
-- Version Link: the algorithm will prioritize sending traffic to instances that have the latest version
+### Algorithm improvement
```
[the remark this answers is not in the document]
aleortega: Added here -> [feedback: list current rules in place](https://github.com/decentraland/adr/pull/220/commits/f0d0a3836be7d2b4e895f679e8abb9c373376d3b)

content/ADR-145-pick-realm-v2.md
aleortega: IMO this is not needed. The capacity itself is configured through environment variables, so it is not really the hard-cap limited by the catalyst resources (_which would be the 100%_). Having said so, there is no a real 100% but the number we configured which can be 300/400/N, before deploying this update to production the default was 2000.

content/ADR-145-pick-realm-v2.md
aleortega: All the pseudo-codes were removed until we have the final implementation in the proposed service. I'll update this ADR to add the pseudocode for each rule once we have each implementation in place to avoid re-work.

Done here -> [feedback: remove pseudo-codes](https://github.com/decentraland/adr/pull/220/commits/1891f6ba8532b35fc25585c718f749f4f523bff1)

content/ADR-145-pick-realm-v2.md
```diff
@@ -9,96 +9,170 @@ spdx-license: CC0-1.0
...
+
+The algorithm will be implemented as a serverless service, outside of the Catalyst services bundle, to accommodate various client implementations. While this may involve modifying the `LARGE_LATENCY` rule to obtain necessary information from the client-side, it opens up possibilities for enhancements such as implementing cache mechanisms or event-driven notifications to proactively filter out unhealthy Catalysts, leading to improved algorithm performance. For further details on this specific rule implementation, please refer to the section on the Large Latency rule.
+
+To fulfill the goal of creating a cost-effective, scalable and globally accessible service, two implementation proposals, CloudFlare Workers and AWS Lambdas, are being considered. The optimal choice for this service will be determined by analyzing various aspects related to cost, pricing model, performance, and latency, which will be discussed in detail in the Server-Side Implementation section.
+
+To assess these aspects, we will examine the workload of the current algorithm, which is monitored through Segment events. The chart below presents the workload data for the past 4 months as of the time of writing.
```
agusaldasoro, line 62:
```suggestion
To assess these aspects, the workload of the current algorithm will be examined, which is monitored through Segment events. The chart below presents the workload data for the past 4 months as of the time of writing.
```

agusaldasoro (approved):
Great work!

agusaldasoro:
> Hello, just for your information I'm delaying the update of this ADR until we have it in place on server-side so I can fulfill the implementation details asked on this PR.

Does it make sense to merge it in "Draft" and update the missing parts in another PR?

aleortega:
The implementation of this ADR has been discarded due to the prioritization of [Default realm](https://governance.decentraland.org/proposal/?id=4c64c0f0-1f20-11ee-9bcb-ff4f8c446533).

PR #232, opened 2023-05-09
docs: new profile validation to prevent repeated wearables

marianogoldman:


content/ADR-XX-validate-repeated-wearables.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Proposed solution
+
+Add a validation to be run on profile deployment that prevents repeated
+wearables from being submitted.
+
+## Deadline
+
+    ADR232_DEADLINE: 2023-06-12T12:00:00Z
+    Unix Timestamp: 1685126400000
```
menduz, line 29: I think we should NOT add this validation. The schema is completely valid and it is an array. This should be only a client side validation

marianogoldman: What would be the purpose for allowing multiple times the same wearable?

menduz (approved):

PR #240, opened 2023-05-23
add ui components

gonpombo8:


content/ADR-237-ui-custom-components.md
```diff
@@ -0,0 +1,144 @@
...
+Now, you can use the `useContext` hook in any functional component to access the current theme and the `toggleTheme` function:
+
+```tsx
+import ReactEcs, { UiEntity } from '@dcl/sdk/react-ecs'
+import React from 'react'
+import { Color4 } from '@dcl/sdk/math'
+
+import { ThemeContext } from './ThemeContext'
+
+const Button = () => {
```
menduz, line 98: Since this will be used as reference, I'd add the `React.FC` types here and wherever possible

menduz (approved):
amazing, maybe linking to the react style rules from dapps may become handy for readers

content/ADR-237-ui-custom-components.md
gonpombo8: I think i have to export our own FC types because they conflict with the ones with react 😞

PR #246, opened 2023-07-26
feat: POC Presence

agusaldasoro:
ADR that describes the scope of the PoC for migrating Presence feature from Matrix to Social Service

### DIFF
--- a/content/ADR-246-poc-presence.md
+++ b/content/ADR-246-poc-presence.md
@@ -32,7 +32,7 @@ This is a PoC so handling different status of presence is out of the scope, the 
 
 ## Specification
 
-The current authentication flow relays on the client to obtain the Matrix Token by themselves, so the ideal will be that when migrating the Presence Feature from Synapse to the Social service implement the Authentication Flow in the Social Service as discussed in [ADR-143](https://adr.decentraland.org/adr/ADR-143).
+The current authentication flow relies on the client to obtain the Matrix Token by themselves, so the ideal would be that when migrating the Presence Feature from Synapse to the Social service, the Authentication Flow should be implemented in the Social Service as discussed in [ADR-143](https://adr.decentraland.org/adr/ADR-143).
 
 As this is a PoC, the focus will be unblocking the project and reduce any risk that can be found. In that line, the PoC doesn't include any change on the Authentication Flow and the current one is the one that's going to be used.
 
--- a/content/ADR-246-poc-presence.md
+++ b/content/ADR-246-poc-presence.md
@@ -18,38 +18,39 @@ The new system will be more reliable, efficient, and scalable, and it will allow
 
 This documents presents a PoC of the implementation for the Presence Notifications in Social Service.
 
-
 ## Context, Reach & Prioritization
 
 Doing this PoC is necessary to understand the effort estimation of migrating Presence feature from Synapse to the Social Service.
 
+The feature needed is to broadcast to all user's friends the position, realm and status (online/offline).
 
 ## Solution Space Exploration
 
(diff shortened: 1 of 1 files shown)


content/ADR-246-poc-presence.md
```diff
@@ -0,0 +1,56 @@
...
+rpc SetPresenceStatus(StatusPayload) returns (StatusResponse) {}
+rpc GetOnlineFriends(Payload) returns (stream UsersResponse) {}
+rpc SubscribeFriendshipPresenceUpdates(Payload) returns (stream SubscribeFriendshipPresenceUpdatesResponse) {}
+```
+
+As the current social service doesn't implement itself authentication, there is no way to automatically identify online users. Doing that is the correct solution, but it's out of scope for this PoC.
+
+So, the proposed solution here isthat when a user wants to be set as active in presence, then they will send the message `SetPresenceStatus` with the status online, so it will be visible. The same if they want to appear offline.
+
+Note that marking a user as disconnected when loosing the connection to the server is out of the scope too, as for that it's needed to have a way to identify the user address from the connection.
```
moliva, line 47: Wouldn't we already have this information? My assumption was that this was using the same auth mechanism as the rest of the Social Service for friendship, allowing us to associate a user (wallet) with the transport.

content/ADR-246-poc-presence.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+rpc SetPresenceStatus(StatusPayload) returns (StatusResponse) {}
+rpc GetOnlineFriends(Payload) returns (stream UsersResponse) {}
+rpc SubscribeFriendshipPresenceUpdates(Payload) returns (stream SubscribeFriendshipPresenceUpdatesResponse) {}
+```
+
+As the current social service doesn't implement itself authentication, there is no way to automatically identify online users. Doing that is the correct solution, but it's out of scope for this PoC.
+
+So, the proposed solution here isthat when a user wants to be set as active in presence, then they will send the message `SetPresenceStatus` with the status online, so it will be visible. The same if they want to appear offline.
```
moliva, line 45: Note that for a migration of the current functionality, users should not only be able to set their status, but along it changes in their position and realm/world.

content/ADR-246-poc-presence.md
agusaldasoro: You are right, I added more details in the authentication mechanism and also added the offline trigger when a user stops sending heartbeats

content/ADR-246-poc-presence.md
agusaldasoro: True! I forgot to add the position, added in the new version!

content/ADR-246-poc-presence.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Solution Space Exploration
+
+Use Redis to store all the connected users, so every time every other user connects to the social service they can retrieve all the friends that are online. Then they can subscribe to all status updates so know when a friend connects or disconnects.
+
+This is a PoC so handling different status of presence is out of the scope, the idea is to migrate the same functionality that currently Matrix has to the Social Server.
+
+## Specification
+
+The current authentication flow relays on the client to obtain the Matrix Token by themselves, so the ideal will be that when migrating the Presence Feature from Synapse to the Social service implement the Authentication Flow in the Social Service as discussed in [ADR-143](https://adr.decentraland.org/adr/ADR-143).
```
Julieta11, line 35: just a tiny typo and nits 
```suggestion
The current authentication flow relies on the client to obtain the Matrix Token by themselves, so the ideal would be that when migrating the Presence Feature from Synapse to the Social service, the Authentication Flow should be implemented in the Social Service as discussed in [ADR-143](https://adr.decentraland.org/adr/ADR-143).
```

LautaroPetaccio (approved):

PR #270, opened 2024-09-28
chore: add example code to adr 239

leanmendoza:
While updating the `wearable-preview`, I had trouble finding the correct way to apply this ADR.
This example code illustrates how the bodyShape ultimately uses the hidden part of the visible wearables, and not reusing the already computed list.

### DIFF
--- a/content/ADR-239-avatar-assembly-changes.md
+++ b/content/ADR-239-avatar-assembly-changes.md
@@ -194,3 +194,44 @@ With the points previously discovered, the suggested procedure to visualize an a
 * Request the wearables listed in the profile wearables field, all wearables listed are the equipped ones, no matter their visibility. ( `POST https://content-server/content/active {"pointers": ["urn1", "urn2"]}` )
 * Process the hide list provided in each wearable following the visualization priority ( the field is the array `data.hides` in each wearable definition )
 * Apply any force render category listed in the forceRender array in the profile definition ( obtained in the profile response called `forceRender` )
+
+### Example code
+```typescript
+
+function getHiddenList(profile: UserProfile): HideableCategory[] {
+   const alreadyHidden = new Set<HideableWearableCategory>()
+   const wearables: Map<WearableCategory, Wearable> = profile.wearables
+   for (const category of categoriesPriority) {
+      const wearable = wearables.get(category)
+      if (!wearable) {
+         continue
+      }
+
+      if (alreadyHidden.has(category)) {
+         continue
+      }
+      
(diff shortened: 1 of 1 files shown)


kuruk-mm (approved):

content/ADR-239-avatar-assembly-changes.md
```diff
@@ -194,3 +194,44 @@ With the points previously discovered, the suggested procedure to visualize an a
...
+const toHide = getHiddenList(profile)
+
+// 2) Remove the hidden wearables
+for (const category of toHide) {
+   wearables.delete(category)
+}
+
+// 3) Compute the current hidden list for hiding parts of the bodyshape
+const hidden = getHiddenList(profile)
+const bodyShape = getBodyShape(profile, hidden)
```
cazala, line 234: What is the difference between `toHide` and `hidden` here?

leanmendoza: I've just fixed a parameter for the `hidden` one.
- `toHide` is the categories that are going to be hide **in the wearable list**
- given the already known wearables are in the list, `hidden` is the categories union of hides of those wearable list, and also applies for the bodyShape, at this point, wearable list is immutable

With this consideration, forceRender is a parameter that once you have a wearable list, you can go to pick one by one the visibility, but force-rendering a specific category is not going to change the `toHide` list (only by the element you pick), on the `hidden` list hand, it might change because it's assuming the entire wearable list is already visible

content/ADR-239-avatar-assembly-changes.md
```diff
@@ -194,3 +194,51 @@ With the points previously discovered, the suggested procedure to visualize an a
 * Request the wearables listed in the profile wearables field, all wearables listed are the equipped ones, no matter their visibility. ( `POST https://content-server/content/active {"pointers": ["urn1", "urn2"]}` )
 * Process the hide list provided in each wearable following the visualization priority ( the field is the array `data.hides` in each wearable definition )
 * Apply any force render category listed in the forceRender array in the profile definition ( obtained in the profile response called `forceRender` )
+
+### Example code
+```typescript
+
+/** 
+ * @param profile - holds the wearables equipped, bodyShape and the forceRender parameter
+ * @param forceAll - wheter if ALL the hides in wearables equipped should be considered or not
```
cazala, line 203:
```suggestion
 * @param forceAll - whether ALL the hides in wearables equipped should be considered or not 
```

cazala (approved):

LautaroPetaccio (approved):

PR #276, opened 2024-11-13
light soruces

nearnshaw:


content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Component description
+
+We should create a `LightSource` component that, when added to an entity, tells the engine to shine a light from that entity’s Transform position.
+
+A Type field will let you chose between _Spot_ and _Point_ light. We believe these two types of lights are enough for now, the component could be expanded in the future if we want to include other types.
+
+The following fields will be available on both types of light:
+
+- Color: _Color4_
```
aixaCode, line 42: Maybe we can provide description of what each filed for with short description, and also limitation on what values it can have

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,116 @@
...
+- Intensity: _number_
+- Range: _number_
+- Active: _boolean_
+
+In lights of type _Spot_, we will also include:
+
+- Inner angle
+- Outer angle
```
aixaCode, line 50: Here we could include the restriction with 

> max 180 - inner angle can’t be more than outer angle

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,116 @@
+---
+layout: adr
+adr: 253
+title: Light Sources
+date: 2024-11-13
+status: Review
```
aixaCode, line 6: Probably at this stage this should be a draft

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,116 @@
...
+
+To be able to play with darkness, we'll also need to provide controls to be able to disable the global default light sources and play with that.
+
+Note: Some GLTF models come with lights packaged as nodes inside the structure of the model. We should ignore those completely. The only lights we will be rendering are the ones defined via code using the LightSource component.
+
+## Component description
+
+We should create a `LightSource` component that, when added to an entity, tells the engine to shine a light from that entity’s Transform position.
+
+A Type field will let you chose between _Spot_ and _Point_ light. We believe these two types of lights are enough for now, the component could be expanded in the future if we want to include other types.
```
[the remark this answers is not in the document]
nearnshaw: Addressed on last commit

content/ADR-253-light-sources.md
nearnshaw: Addressed on last commit

content/ADR-253-light-sources.md
nearnshaw: Addressed on last commit

content/ADR-253-light-sources.md
nearnshaw: Addressed on last commit

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,117 @@
...
+
+A _Spot_ light projects a cone of concentrated light in a single direction, originating from a single point in the scene. It's ideal for shows and theatric effects, it can also be used to shine a light on something with a functional intention.
+
+A _Point_ light expands in all directions from a point in the scene, illuminating the space in a more subtle way. The point of origin is often not easy to pinpoint for the player.
+
+The following fields will be available on both types of light:
+
+- Color: _Color4_ The color of the light
+- Intensity: _number_ The luminosity value of the light, from 0 to 1.
+- Range: _number_ The maximum distance that can be affected by the light source.
```
leanmendoza, line 48: Regarding light range, I suggest following GLTF's [KHR_lights_punctual](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_lights_punctual/README.md) approach:

1. The range should be **optional** (or even more not being added now). When unspecified, light naturally SHOULD attenuate following the inverse square law (1/d²)

2. When specified, GLTF uses this attenuation formula after following the inverse square law, then:
```glsl
attenuation = max(min(1.0 - (distance/range)^4, 1), 0) / distance^2
```
`Within the range of the light, attenuation should follow the inverse square law as closely as possible, although some non-quadratic falloff near the edge of the range may be used to avoid a hard cutoff`

This provides:
- Natural physical falloff (1/d²)
- Smooth cutoff at the range distance
- Performance optimization opportunity (GPU can skip calculations beyond range)

Instead of having developers guess appropriate range values, we could:
- Make range optional
- Calculate a reasonable default cutoff based on illuminance
- Allow manual override when needed for performance tuning

This aligns with industry standards while maintaining physically-based behavior.

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,117 @@
...
+
+To be able to play with darkness, we'll also need to provide controls to be able to disable the global default light sources and play with that.
+
+Note: Some GLTF models come with lights packaged as nodes inside the structure of the model. We should ignore those completely. The only lights we will be rendering are the ones defined via code using the LightSource component.
+
+## Component description
+
+We should create a `LightSource` component that, when added to an entity, tells the engine to shine a light from that entity’s Transform position.
+
+A Type field will let you chose between _Spot_ and _Point_ light. We believe these two types of lights are enough for now, the component could be expanded in the future if we want to include other types.
```
leanmendoza, line 38: Regarding the component structure, I'd like to discuss an alternative approach based on composition rather than a single component with type discrimination:

Instead of:
```typescript
interface LightSource {
  type: 'point' | 'spot' | 'directional';
  intensity: number;
  // ... common properties ...
  // spot-specific properties when type === 'spot'
  angle?: number;
  innerAngle?: number;
}
```

Consider a composable approach:
```typescript
interface Light {
  illuminance: number;
  color: Color3;
  shadows: boolean;
}

interface Spotlight {
  angle: number;
  innerAngle?: number;
}

interface GlobalLight {
  direction?: Vector3;
  ambientColor?: Color3;
  ambientBrightness?: number;
}
```

Benefits:
1. **Cleaner Separation**: Each component handles one specific aspect of lighting
2. **Type Safety**: No need for runtime type checking or optional properties
3. **Extensibility**: New light types can be added without modifying existing components
4. **Flexibility**: SDKs can provide their own abstractions while the protocol remains clean
5. **Clarity**: Entity behavior is determined by component composition:
   - Point Light = Entity + Light
   - Spot Light = Entity + Light + Spotlight
   - Directional Light = Root Entity + Light + GlobalLight

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,117 @@
...
+- Active: _boolean_ Determines if the light is currently on or not.
+
+In lights of type _Spot_, we will also include:
+
+- Inner angle: _number_ This angle, measured from the direction vertex, defines a cone where the light has full intensity. Max 180, can’t be more than outer angle.
+- Outer angle: _number_ This angle, measured from the direction vertex, defines a cone where the light has an effect, the intensity decreases farther away from the inner cone. Max 180.
+
+The `Active` flag lets creators easily turn a light source on or off. We could otherwise achieve the same by setting intensity to 0, but offering a switch makes it easier to retain the prior configuration.
+
+## Shadows
```
leanmendoza, line 58: Regarding shadows, I suggest expanding this section to address several important aspects:

1. **Shadow Control**
Instead of a simple boolean, consider these key properties:
- `shadows: boolean` - Whether the light casts shadows
- Default values should differ by light type:
  * Point/Spot lights: `false` (performance consideration)
  * Global/Directional light: `true` (essential for scene depth)

2. **Implementation Considerations**
Add a note that shadow rendering may vary by platform:
```typescript
// Even when shadows = true, the engine may:
// - Not display shadows at all
// - Show shadows for limited number of lights
// - Vary shadow quality
// Based on:
// - Platform capabilities
// - User settings
// - Performance requirements
```

3. **Priority System**
Consider adding guidance for implementations about shadow priority:
- Global/Directional light shadows take precedence
- Point/Spot lights could use distance/illuminance for priority

This approach:
- Gives developers clear expectations
- Allows implementations to optimize performance
- Maintains flexibility for different platforms
- Provides predictable behavior across scenes

See our protocol implementation for a working example of these concepts.

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,117 @@
...
+## Open world considerations
+
+We advise that each engine only renders the lights of the current scene you’re standing on. In this way:
+
+- Neighboring scenes don’t get affected in performance
+- Neighboring scenes don’t mess with the esthetic of others, at least not when you’re standing on that other scene.
+
+Engines can add a default behavior of fading lights in/out over a period of a second whenever you switch scenes. This is to avoid abrupt light changes when you walk from one parcel to another.
+
+### Affect global ambient light
```
leanmendoza, line 93: The Global Light section needs more detailed specification. Currently, there's a gap between point/spot lights and scene-wide lighting. Consider:

1. **Global Light Components**
A scene's global lighting should be configurable through two complementary components:

```typescript
// Core light properties (when attached to root)
interface Light {
  illuminance: number;  // 400 (sunrise) to 10,000 (midday)
  color: Color3;
  shadows: boolean;
}

// Global light specific properties
interface GlobalLight {
  direction?: Vector3;      // Sunlight direction
  ambientColor?: Color3;    // Sky/environment contribution
  ambientBrightness?: number; // Overall scene brightness multiplier
}
```

2. **Root Entity Behavior**
- `Light` on root = Override default directional (sun) light
- `GlobalLight` on root = Control direction and ambient settings
- Both components can work independently:
  * `Light` alone = Change sun color/intensity
  * `GlobalLight` alone = Change direction/ambient
  * Both = Full control

3. **Default Behavior**
Add specification for default values:
- Default directional light should match typical daylight
- Direction could vary with time-of-day (if implemented)
- Reasonable ambient light for scene visibility

This matches our protocol implementation while providing clear guidance for scene creators and engine implementations.

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,117 @@
...
+By default lights won’t have shadows. Each light source can chose if they want shadows or not, and if to use hard shadows or soft shadows.
+
+We will add fields for this on the `LightSource` component:
+
+- Shadow type: No shadows / Hard Shadows / Soft shadows
+
+- The creator can chose the shadow resolution as a setting on each light source
+- The shadow resolution is picked by the player in the user’s settings, as a global option. If they have low settings they’ll always use low res
+
+## Limitations
```
leanmendoza, line 71: I suggest removing the "Limitations and Considerations" section from this ADR. Here's why:

1. **Separation of Concerns**
- This ADR should focus on defining the light component interfaces and their behavior
- Implementation details like performance limits should be handled separately:
  * Engine-specific optimizations
  * Platform-specific limitations
  * Scene-level restrictions 

2. **Flexibility for Implementations**
Let each explorer/renderer decide how to handle:
- Maximum number of lights
- Shadow map allocation
- Performance optimizations
- Mobile vs desktop capabilities
- Low-end vs high-end configurations

3. **SDK's Role**
The SDK can provide:
- Platform-specific warnings
- Best practice guidelines
- Performance recommendations
- Scene validation tools

4. **Better Documentation Structure**
These topics deserve their own documentation:
- Performance best practices ADR
- Scene optimization guide
- Platform compatibility matrix
- Implementation guidelines

This approach:
- Keeps the component specification clean and focused
- Allows for platform-specific optimizations
- Enables future improvements without spec changes
- Lets implementations evolve independently

leanmendoza (changes requested):
Thank you for putting this ADR for light sources! This is a crucial feature for Decentraland. 
I want to take this place to debate this, given the previous work we've done in the protocol. Please take a look before reading the next part:
- https://github.com/decentraland/protocol/blob/protocol-squad/proto/decentraland/sdk/components/light.proto
- https://github.com/decentraland/protocol/blob/protocol-squad/proto/decentraland/sdk/components/spotlight.proto
- https://github.com/decentraland/protocol/blob/protocol-squad/proto/decentraland/sdk/components/global_light.proto

I had written a whole text to comment as a review, but then I split into different topics and comment each part of the ADR to keep the conversation cleaner :)

content/ADR-253-light-sources.md
```diff
@@ -0,0 +1,117 @@
...
+A Type field will let you chose between _Spot_ and _Point_ light. We believe these two types of lights are enough for now, the component could be expanded in the future if we want to include other types.
+
+A _Spot_ light projects a cone of concentrated light in a single direction, originating from a single point in the scene. It's ideal for shows and theatric effects, it can also be used to shine a light on something with a functional intention.
+
+A _Point_ light expands in all directions from a point in the scene, illuminating the space in a more subtle way. The point of origin is often not easy to pinpoint for the player.
+
+The following fields will be available on both types of light:
+
+- Color: _Color4_ The color of the light
+- Intensity: _number_ The luminosity value of the light, from 0 to 1.
```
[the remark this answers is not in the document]
aixaCode: Good suggestion, how about we use Lumens instead, it measure the total amount of light emitted by a source. This makes it the ideal metric for describing the output of a light source, independent of where or how it is used. Lumens are inherently tied to the light source itself, while lux depends on the environment. Since we want to define properties of the light sources rather than the effect of light on a surface, lumens are a better fit, in my option.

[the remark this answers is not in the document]
leanmendoza: We could use Lumens of course, that would be equivalent to divide the **luxs** by **4*π**. But you should keep using **lux** for directional lights since the source is practically infinity. And this differentiation is the one glTF extension uses.

glTF extension:
intensity | Brightness of light in. The units that this is defined in depend on the type of light. point and spot lights use luminous intensity in candela (lm/sr) while directional lights use illuminance in lux (lm/m2)
-- | --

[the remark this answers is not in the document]
nearnshaw: After talking it over and looking at other platforms, we changed the value to be measured in lumens and renamed to "brightness"

content/ADR-253-light-sources.md
nearnshaw: We debated this point, and reached the conclusion that we still much prefer to use a single component, for the following reasons:

1. It's consistent with how many other components in our SDK already work (eg Material, Tween, MeshRenderer, MeshCollider)
2. It's friendlier to have to remember a single component name, and have the field suggestions do the rest for you. If you need to add two separate components, you don't have that guidance, you need to remember the two component names by heart. Users can also get potentially lost if they try to only add the SpotLight component and see no effect.
3. We can add helpers, like we already do on other components to make this friendlier. eg:
`LightSource.spotLight(myEntity, { brightness: 1000, angle: 10})`

content/ADR-253-light-sources.md
nearnshaw: We agree that it's better to have shadows as a simple boolean, and let each engine determine things like shadow resolution, priorities, etc.
I changed this in the document

content/ADR-253-light-sources.md
nearnshaw: Agreed, I expanded the section, and added more to the initial definition. I however added this as an additional _type_, rather than its own component.

content/ADR-253-light-sources.md
nearnshaw: I agree with both, the ADR should focus on the protocol, but sometimes the protocol is more than just the data sent via the transport, it's also about the expected behavior.

I simplified these lines to something more generic:

> Each engine is free to determine considerations like shadow resolutions, or putting a limit on the number of shadows being computed and how to prioritize these. It's recommendable to make these variables dependent on user quality settings.

content/ADR-253-light-sources.md
nearnshaw: We agree that the attenuation curve of light should mimic real life as closely as possible
The "range" field should serve as a hard cut-off, but not distort the attenuation curve.

It's mostly there to allow creators to optimize. The default value should be something reasonable, and most novice creators might not need to tweak it

content/ADR-253-light-sources.md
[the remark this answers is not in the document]
leanmendoza: And luxs for directional/global?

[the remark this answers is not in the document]
nearnshaw: You're right, I was leaving that out bc global lights aren't in the plan for our first implementation, but it should be in the ADR.
I just included a clarification for that: 

> For _Spot_ and _Point_ light, this value is expressed in Lumens, for _Global_ light, it's expressed in Lux units (lumens per square metre).

content/ADR-253-light-sources.md
leanmendoza: So, do we agree with letting the `attenuation curve` be fixed to the inverse square law? And then, the range is just a cutoff and is up to implementation if this is a step function or something smoother

nearnshaw: Yes, we agree on this point :)

PR #280, opened 2025-01-03
ADR:257 Avatar Rotation When Moving Player

nearnshaw:


leanmendoza (approved):
Is it the same as changing `Transform.getMutable(engine.PlayerEntity).rotation`, right?
LGTM!

nearnshaw:
> Is it the same as changing `Transform.getMutable(engine.PlayerEntity).rotation`, right? LGTM!

The result is the same, yes
The difference is that instead of setting it as a Quaternion value, we're setting it as a Vector3 for a position in the scene space that the avatar needs to face. In the ADR I go into detail about why

aixaCode (approved):

PR #281, opened 2025-01-03
ADR 256: static entities

nearnshaw:


content/ADR-256-static-entities.md
```diff
@@ -0,0 +1,50 @@
...
+
+All static entities should be instanced on the scene's first frame, either as part of a `.composite` file (that is created by interacting with the Creator Hub UI) or instanced on the `main()` function. A static entity should remain static for all of the scene's lifecycle, it can't be removed or switched back to non-static, since several engine optimizations might be relying on this state being permanent.
+
+Note: As an alternative approach, we also discussed handling this via a separate new `Static` component instead of a new field on the `Transform` component. There are pros and cons to each approach.
+
+- On one hand, an extra field on the Transform adds a slight overhead to each update message regarding any transform. (Although if the value of its corresponding protobuff digit is 1, then there should be no Transform updates beyond the initial creation)
+- On the other hand, it's easier to creators to find where to add this property if it exists on the Transform. Any object that makes sense to make static will always have a Transform anyway.
```
NickKhalow, line 46: I'd choose Static component rather than an extra bool in Transform.
1. Static feature affect many other components beside Transform, it makes more sense to implement it as a new component in my opinion
2. I'd say that it would affect L-caching performance on engine's side because each extra field makes an array of structures bigger and it turns to handle less components at the time. But this aspect could be easily optimised

kuruk-mm:
Adding a field to the Transform breaks the compatibility of the protocol. The Transform is not a protobuf; it's a fixed set of bytes. Adding a new field changes its length, which makes it problematic. Additionally, it's the most used component. Adding one byte that isn't going to be used is not a good option.

Optionally, we can use a new `StaticComponent`. To make things easier for Creators, we could create a wrapper in the SDK that behaves like a new field in the Transform but adds the `StaticComponent` behind the scenes.

Almost every time we make a protocol change to "make things easier for users," the solution doesn’t lie in the protocol itself—it lies in the SDK implementation. The protocol should remain as simple as possible, serving as an efficient communication channel between the runtime and the explorers.

Anyway, we have another proposal for this approach that we think is better for the Creator and also easier for the Protocol: adding this function to the `engine` API.

Example code:
```typescript
import { markEntitiesAsStatic } from '~system/EngineApi'

export function main() {
  const door = engine.addEntity()

  GltfContainer.create(door)
  markEntitiesAsStatic([door])

  GltfContainer.update(door) // it should fail at the SDK level... at the engine level, it does nothing
}
```

WDYT?

Also. It would be interesting to know which kind of optimization the Foundation is thinking.

content/ADR-256-static-entities.md
AlejandroAlvarezMelucciDCL: My 2 cents:
1. Yes it does, but all the components that are affected have a dependency with Transform, just like in Unity Editor, the static check is in the GameObject root and not in every component, you can't mark a component static and not the root GameObject and all GameObject have a Transform by default that's not removable. If we're talking about SDK components that can be "static" without Transform, I think that's another property of that component, not directly tied to this static concept discussed here.
2. Protocol properties marked as optional are ignored/removed if not set, so I don't think the performance impact would be that measurable, we should run some tests and gather some data before deciding on that alone.

PR #290, opened 2025-02-19
trigger areas

nearnshaw:


robtfm (commented):
- we should ensure that triggers are triggered even if the object "passes through" the target in a frame. this would require special handling for tweens. or explicitly say that this doesn't trigger (which would be a shame)
- we should report the position of the trigger as well (at first point of contact) for the case when the entity with the trigger has a tween
- if an entity has collision=physics and trigger=player set to the same mesh, should we fire the trigger when player touches? from the example it appears so, this should probably be explicitly mentioned either way

content/ADR-258-trigger-areas.md
```diff
@@ -0,0 +1,165 @@
...
+
+- `triggeredEntity`: The entity that was triggered (this is the entity that owns the trigger area)
+- `state`: The state of the trigger event (ENTER, EXIT, STAY)
+- `timestamp`: The timestamp of the trigger event
+- `trigger`: An object with the following fields:
+  - `entity`: The entity that triggered the trigger
+  - `layer`: The collision layer of the entity that triggered the trigger (this is the same as the trigger layer of the trigger area)
+  - `position`: The position of the entity that triggered the trigger
+  - `rotation`: The rotation of the entity that triggered the trigger
+  - `scale`: The scale of the entity that triggered the trigger
```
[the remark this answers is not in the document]
nearnshaw: The same would apply to the fields `triggeredEntityPosition` and `triggeredEntityRotation` that I just added
If we change the 3 above fields for a Transform, then we should do the same with the triggered entity

kuruk-mm (approved):
Great job 🙌

content/ADR-258-trigger-areas.md
```diff
@@ -0,0 +1,193 @@
...
+
+When using the `PHYSICS` layer, the trigger area will trigger events when any entity with a physics collider overlaps with it.
+
+Note: If a collider with the same shape blocks a player from entering a trigger area, then we won't consider a trigger event being sent. There must be an overlap with the area, not just contact. For example, if a same entity has both a MeshCollider component set to the layer `CL_PHYSICS` and a Trigger component set to the layer `CL_PLAYER`, and both meshes have the same shape, then the player will never be able to activate a trigger event from this entity. In the future we'll likely want to do _Collision Events_ as a separate feature.
+
+### Trigger events
+
+Trigger areas can trigger events when the player (or any other entity on the trigger layer) enters, exits or stays in the area.
+
+Trigger events would have to be shared from the engine to the SDK via a component, following a similar approach as we do with pointer events and raycasts. We should create a `TriggerCollisionResult` component. Creators are not expected to read values or make use of this component in any way, unless they really want to fine tune their scene’s behavior.
```
pravusjif, line 76: Suggesting based on what we ended up implementing for [TriggerAreaResult component](https://github.com/decentraland/protocol/blob/258fabc1584701e63f129af3ca9895960305f703/proto/decentraland/sdk/components/trigger_area_result.proto):

```suggestion
Trigger events would have to be shared from the engine to the SDK via a component, following a similar approach as we do with pointer events and raycasts. We should create a `TriggerAreaResult` component. Creators are not expected to read values or make use of this component in any way, unless they really want to fine tune their scene’s behavior.
```

content/ADR-258-trigger-areas.md
```diff
@@ -0,0 +1,193 @@
...
+
+This component will be added by the engine, similarly to how pointer events are handled. The event will be triggered when the entity enters, exits or stays in the area. Each event will have the following fields:
+
+- `triggeredEntity`: The entity that was triggered (this is the entity that owns the trigger area)
+- `triggeredEntityPosition`: The position of the triggered entity at the time of the trigger.
+- `triggeredEntityRotation`: The rotation of the triggered entity at the time of the trigger.
+- `state`: The state of the trigger event (ENTER, EXIT, STAY)
```
pravusjif, line 83: Suggesting based on what we ended up implementing for [TriggerAreaResult component](https://github.com/decentraland/protocol/blob/258fabc1584701e63f129af3ca9895960305f703/proto/decentraland/sdk/components/trigger_area_result.proto):

```suggestion
- `eventType`: The state of the trigger event (ENTER, EXIT, STAY)
```

content/ADR-258-trigger-areas.md
```diff
@@ -0,0 +1,193 @@
...
+- `timestamp`: The timestamp of the trigger event
+- `trigger`: An object with the following fields:
+  - `entity`: The entity that triggered the trigger
+  - `layer`: The collision layer of the entity that triggered the trigger (this is the same as the trigger layer of the trigger area)
```
pravusjif, line 87: Suggesting based on what we ended up implementing for [TriggerAreaResult component](https://github.com/decentraland/protocol/blob/258fabc1584701e63f129af3ca9895960305f703/proto/decentraland/sdk/components/trigger_area_result.proto):
```suggestion
  - `layers`: The collision layermask of the entity that triggered the Trigger Area
```

content/ADR-258-trigger-areas.md
```diff
@@ -0,0 +1,193 @@
...
+      entity: myTrigger,
+      opts: {
+        layer: Player
+      }
+    },
+    function (otherEntity) {
+      // Do whatever I want
+    }
+  )
+```
```
pravusjif, line 122: Suggesting based on what we ended up implementing [in the SDK](http://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/ecs/src/systems/triggerArea.ts):

```suggestion
```ts
TriggerArea.setBox(myTrigger)
Transform.create(myTrigger, {
  position: Vector3.create(8, 0, 8),
  scale: Vector3.create(1, 1, 1),
})
triggerAreaEventsSystem.onTriggerEnter(myTrigger, function(result) {
  // Do whatever I want
})
```
```

pravusjif (approved):

PR #293, opened 2025-03-13
ADR 259: skybox time

nearnshaw:


content/ADR-259-skybox-time.md
```diff
@@ -0,0 +1,63 @@
...
+status: Draft
+type: RFC
+spdx-license: CC0-1.0
+authors:
+  - nearnshaw
+---
+
+## Abstract
+
+The current protocol supports allowing a scene to have a fixed time of day, but it doesn't allow the creators to change it on the fly.
```
pravusjif, line 15: Should we mention that's only available inside the `"worldConfiguration"` structure in the scene.json? to make clear it's only available for worlds today.

content/ADR-259-skybox-time.md
```diff
@@ -0,0 +1,63 @@
...
+
+Tip: To interpret to this value, divide the seconds value by 60 to obtain minutes, and by 60 again to obtain the hours since the start of the day. For example, if the seconds value is 36000, it corresponds to 10:00 am.
+
+This form of expressing the time of day is consistent with the way we express the `fixedTime` on the scene.json file.
+
+If this component is not present, the scene will follow the same day/night cycle as Genesis City.
+
+## Transitioning time
+
+Whenever the `time` property changes, the scene will transition to the new time of day over a period of time to avoid abrupt changes. The details about this transition are to be defined by each engine. We're intentionally not exposing a `transitionTime` property in this component, as we want to provide that control to the engine.
```
pravusjif, line 36: however, there could be another property like `transitionMode` allowing 2 possible modes: `forward`, `backward`.

Othwerwise, imagine if you are at let's say `Skyboxtime.time = 36000` and you want to now transition to be at `30000`, should the time transition forward until reaching the desired value? or transition backward to that time because it's "closer" to the current time?

I think it would be nice to give control to creators on that.

kuruk-mm: Maybe you can add also `transitionMode: auto / closer / near` (could be calculate by the user, but it's a nice to have there)

nearnshaw: I think the transitionMode can be useful

@kuruk-mm what do you image for the "closer" and "near" values? I don't have a clear idea of what those would do

I would not let the sky switch without a transition, or with a custom duration, since that could cause jarring effects when walking from scene to scene

nearnshaw: Wouldn't the creator be able to read the skybox time from the new `Skyboxtime` component we're creating here?

Maybe I should specify this, but I imagine that it would be possible to read the current time from off to `time` property. Would that work? or would it add extra complexity? I don't think it's worth updating this value on every tick with engine updates if it's not going to be used if that's needed.

pravusjif: The thing is that if the Engine is constantly writing on that component (to constantly update the skybox time on it), when the scene writes on it, I think that change will be lost.

It's the same reason why we can't just write the "current time" from the engine on `Tween` or `VideoPlayer` components, we need the "state components" that the scene can read separately... I think the same will happen here.

pravusjif: > I don't think it's worth updating this value on every tick with engine updates if it's not going to be used if that's needed.

The Engine is already writing on the `EngineInfo` component on every tick

content/ADR-259-skybox-time.md
aixaCode: Is that respected in E@?

content/ADR-259-skybox-time.md
aixaCode: How would that component be working with the global configuration we have right now, that we use for example during music festival? I assume this would replace that? Also for those events, we might want to keep fixed time?

content/ADR-259-skybox-time.md
nearnshaw: This is not currently working on E@, but we plan on implementing this together with the skyboxTime component. It is however already part of the protocol

Would be ideal to accept this property in the scene.json also if not nested inside `worldConfiguration`, so we can have it working straight up in Genesis City scenes too

content/ADR-259-skybox-time.md
nearnshaw: I'm adding a section about hierarchy of definitions to the ADR, to avoid any ambiguity in case of contradicting values.

I'd rank them like this, in order of most weight first:

1. The `SkyboxTime` component.
2. The `fixedTime` property in the scene.json file.
3. The fixed time of day defined by the player via the UI.
4. The time of day as defined globally in Decentraland.

Regarding global night settings, I think I'd leave that outside the protocol discussion, as it's a more engine-specific thing we do on the Unity client. I would put it between 3 & 4, to allow both players and scenes to be able to customize beyond that imposed setting.

nearnshaw: I just pushed some small edits to address the ideas here:

- I added a transitionMode property
- The default direction is forward, unless specified on transitionMode
- I added a section on Hierarchy of day/night definitions

pravusjif (approved):

content/ADR-259-skybox-time.md
kuruk-mm: > I think the transitionMode can be useful to specify the direction of the movement. By default I suppose it should always be the shortest distance.
> 
> @kuruk-mm what do you image for the "closer" and "near" values? I don't have a clear idea of what those would do
> 
> I would not let the sky switch without a transition, or with a custom duration, since that could cause jarring effects when walking from scene to scene

oops I mean `transitionMode: auto | forward | backward`

PR #301, opened 2025-07-31
continuous tweens

nearnshaw:


content/ADR-285-continuous-tweens.md
```diff
@@ -0,0 +1,155 @@
...
+
+The current workaround (chaining multiple finite tweens) introduces complexity for creators and causes visual artifacts due to frame-precision issues and unnatural looping.
+
+## Decision
+
+We will introduce three new continuous tween modes to the Tween component:
+
+- `RotateContinuous`
+- `MoveContinuous`
+- `TextureMoveContinuous`
```
kuruk-mm, line 27: I’d like to start a conversation about an alternative approach—similar to how Unity’s DOTween works.

The idea is to introduce the following new modes to the `Tween` component:

```proto
message MoveIncrement {
  decentraland.common.Vector3 value = 1;
}

message RotateIncrement {
  decentraland.common.Quaternion value = 1;
}

message ScaleIncrement {
  decentraland.common.Vector3 value = 1;
}

message TextureMoveIncrement {
  decentraland.common.Vector2 value = 1;
  optional TextureMovementType movement_type = 2; // default = TextureMovementType.TMT_OFFSET
}
```

In addition, we could introduce a new loop mode for the tween sequence: `TextureLoop.INCREMENTAL`.

Here’s how it would work:

```ts
Tween.create(entity, {
  mode: Tween.Mode.MoveIncrement({ value: Vector3.create(10, 10, 10), duration: 2.0 })
})
```

This would move the entity by 10 units along all axes over 2 seconds—executed once.

If used in a sequence with different loop modes:

* **RESTART**: It moves the entity by 10 units, then immediately resets (subtracts the same amount), and starts over.
* **YOYO**: It moves the entity by 10 units, then reverses by the same amount (like a ping-pong).
* **INCREMENTAL (new)**: It moves the entity by 10 units, and when it ends, it repeats the same increment again.

So, using RESTART and YOYO, you can achieve the same results as with the existing non-incremental modes (START/END), but instead of moving between fixed points, the effect is achieved through value addition and subtraction.

This would still work with the existing `duration` and `easing_function` parameters, and I believe it would be clear and intuitive for creators to use.

What do you think?

kuruk-mm: This ADR should not address that issue; it is focused on introducing new functionality (incremental/continuous tweens).

To resolve the sequence issue, we should push **TweenSequence** into the engine.

menduz (changes requested):

content/ADR-285-continuous-tweens.md
```diff
@@ -0,0 +1,156 @@
+---
+layout: adr
+adr: 285
+title: Continuous tweens
```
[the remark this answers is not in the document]
hukasu: @nearnshaw the link 404

robtfm:
we (bevy team) notice a divergence between unity implementation and adr for continuous tweens with duration. the ADR says that easing should be applied in that circumstance, but unity does not apply any easing on continuous tweens, with or without duration. 

we think unity's approach is more intuitive than the ADR spec, so we implemented the same way.

hukasu:
i mean, having a way to have describe a continuous tween that allows for easing would be useful, like a fan starting from resting and spinning up

PR #302, opened 2025-07-31
voice chat area

nearnshaw:


robtfm:
there are interesting synchronisation and hacking issues here. the existing modifiers only apply to the local client, so it isn't important for the state of the area to match between players. but with this addition we would either delegate all the control of the voice channels to the client (so a malicious client could choose to enter voice areas it should not), or we would want that control to be separated from the client's specific state / managed by a server somehow.

not sure what the answer is for that, just noting it.

content/ADR-286-voice-chat-areas.md
```diff
@@ -0,0 +1,126 @@
...
+status: Living
+type: RFC
+spdx-license: CC0-1.0
+authors:
+  - nearnshaw
+---
+
+# Abstract
+
+Scenes often require localized, ad-hoc voice communication between players who are physically near each other within a specific in-game space — for example, social hubs, games like speed dating, art exhibits, or impromptu stage areas. The current voice system only supports Community Voice Chat (with moderators and persistent rooms) or Private Conversations, which are either too global or too exclusive for these use cases.
```
fcolarich, line 15: Community Voice Chat rooms are not persistent per se, they are automatically discarded if there are no more mods present.

content/ADR-286-voice-chat-areas.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Decision
+
+We are introducing a new type of Avatar Modifier Area called a Voice Chat Area, which automatically places players into a temporary, scoped LiveKit room when they are standing inside a defined 3D region. These areas provide private, ephemeral voice chat spaces where everyone can speak freely without moderation.
```
fcolarich, line 19: The privacy gains are only that the chat is limited to a specific area in the scene? but that can be easily overcome, even if you wall off a scene, you can cheat the client. And as its unmoderated, there is no way to solve that.

content/ADR-286-voice-chat-areas.md
```diff
@@ -0,0 +1,126 @@
...
+
+This ADR specifies the data model, runtime behavior, UI integration, and constraints for implementing Voice Chat Areas.
+
+## Motivation
+
+- Support natural, proximity-based audio conversations within specific scene areas.
+
+- Enable immersive use cases like speed dating, stage performances, open mic corners.
+
+- Avoid eavesdropping and ensure conversations are scoped to what players see.
```
fcolarich, line 37: I dont see how the solution would fix eavesdropping? a scene can have walls and the volumes of the area can easily go beyond walls or over supposed ceilings? there is no guarantee users cannot hide in non-visible parts of the scene? Ideally you would see the list of participants in the call and know who is listening, but that's all we can do, as its unmoderated we cannot remove the eavesdropping users :S 
Also users could capture the connection string and just connect to the call from anywhere, as there is no server validation that you are inside the area when you try to connect. Unless we give really TTL to connection strings, we have no way to handling this. The downside is, if a player disconnects while "legally" in the call, they will need to request a new string as their will have already expired.
And well, as with any unmoderated space, if a user chooses to cause disturbances, there is no way to remediate it.

content/ADR-286-voice-chat-areas.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Exclusivity
+
+- A player cannot be in multiple voice chats simultaneously:
+
+- No overlapping Voice Chat Areas.
+
+- No joining while in a Private or Community Chat.
+
+- Attempts to join while already in a call must trigger a prompt to leave first.
```
fcolarich, line 69: How would joining work? you are joined automatically? or do you need to click a button to join? 
If you are joined automatically, it means as soon as you enter an area with 2 or more people in it already, you will start hearing them, even if you didnt want to.
Or if several players are running through an area and they cross a scene that has a big enough chat area, they will get connected to a room and everything, even thou they have no intention of participating in the chat? wouldn't that be throwing money at the wall?
Regarding this specific point I have another doubt: `Attempts to join while already in a call must trigger a prompt to leave first.` tied to the previous question, are users prompted to join or are they joined automatically? how do you "attempt to join" an area voice chat?

content/ADR-286-voice-chat-areas.md
```diff
@@ -0,0 +1,126 @@
...
+
+- No eavesdropping — players outside the volume cannot hear or join the conversation.
+
+- Participants should feel safe that the conversation is limited to visible, present users.
+
+### Exclusivity
+
+- A player cannot be in multiple voice chats simultaneously:
+
+- No overlapping Voice Chat Areas.
```
fcolarich, line 65: How would this be managed? who decides which areas are valid? when the scene is uploaded? or when its running? how do we ensure all clients make the same decision regarding at which area to ignore?

content/ADR-286-voice-chat-areas.md
```diff
@@ -0,0 +1,126 @@
...
+- Manage joining/leaving LiveKit rooms automatically based on presence.
+
+- Avoid resource waste by not persisting empty rooms.
+
+- Synchronize user presence lists with UI updates in real-time.
+
+
+## Scalability
+
+- Unlimited number of VoiceChatAreas can exist in a scene.
```
fcolarich, line 85: Is this reasonable? scalable? if Livekit charges us for each open room, this an mean a single scene with 100 users spread around, can consume 50 rooms worth of money and if they stay connected, the rooms stay open forever, even if no player is talking.

content/ADR-286-voice-chat-areas.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Only active areas with multiple users consume backend voice resources.
+
+- LiveKit’s built-in user limits (e.g., ~256 per room) apply.
+
+## Excluded IDs
+
+As in other avatar modifier areas, the `excludeIds` field can be used to exclude specific players from the area. This could be useful for creators or eventually scene admins to exclude toxic players from joining the voice conversation.
```
fcolarich, line 93: This doesnt feel like it would solve the problem? as creators would need to re-deploy their scenes for this change to take effect? so if you have an undesirable user, you have to re-deploy and wait until the scene finishes deploying and then join again and hope that undesirable user doesn't join with an alt or you need to re-deploy again 😢

PR #322, opened 2026-09-02
fix: sign the signed-fetch metadata verbatim instead of folding it

LautaroPetaccio:
## Description

ADR-44 specifies the signed-fetch payload as:

```typescript
const payload = [METHOD, PATH, TIMESTAMP, METADATA].join(":").toLowerCase()
```

The fold happens *after* the metadata is joined in, which leaves the metadata's casing **outside** the signature. `{"signer":"decentraland-kernel-scene"}` and `{"Signer":"decentraland-kernel-scene"}` produce a byte-identical payload and therefore share one valid signature — while `X-Identity-Metadata` is delivered exactly as written. A service comparing `metadata.signer` reads the second as *absent*, so a request could be re-spelled in flight, keep a genuinely valid signature, and bypass the check that property was there to enforce.

This amends the specification so the metadata is joined **verbatim**:

```typescript
const payload = [METHOD.toLowerCase(), PATH.toLowerCase(), TIMESTAMP, METADATA].join(":")
```

Method and path are still lowercased, so a signature does not depend on how a client spelled them. Requests are otherwise unchanged — the header carries the same JSON, only the signed string differs.

All three amended documents are status **Living**, which is the status meant to be amended in place rather than superseded.

> This documents behaviour the implementations already have. `@dcl/crypto-middleware` 6.x, `decentraland-crypto-fetch` 3.x, `decentraland-gatsby` 9.x and `dcl-crypto-middleware-rs` 0.3.x all build the payload this way, and the services have been rolled onto them. ADR-44 was the last place still specifying the folded form.

## Changes

**`ADR-44-signed-fetch.md`** — the substantive change.

- The payload now lowercases method and path only and joins the metadata verbatim.
- A new *Metadata is signed verbatim* section explains why, and keeps the superseded form in the document under a `// Superseded. Do not use.` label — implementations still need to recognise it while callers migrate.
- *Notes for implementers* covers what is easy to get wrong: serialize the metadata once and sign the exact bytes sent (key order and whitespace are signed); rebuild the payload from the metadata as delivered and do not normalize it before verifying; compare authorized-on properties exactly rather than case-folding; and if a verifier keeps a fallback to the superseded payload during migration, scope it to the properties it authorizes on and refuse a delivered property name that differs from the expected spelling — because the superseded payload cannot bind it.

**`ADR-289-explorer-client-scenes-signed-fetch.md`** — consistency, not a rewrite.

It states its metadata is part of the signed payload "ensuring its integrity and authenticity", and its fields are camelCase (`sceneId`, `hashPayload`, `realm.serverName`) — precisely what folding left unbound. Its *Security Considerations* covered the client-side trust boundary (scenes cannot reach the metadata the client builds) but not the request once it leaves the client, so a short *Metadata Integrity in Transit* subsection now states that, and the Overview notes the metadata is joined verbatim per ADR-44.

**`ADR-180-communication-adapters.md`** — one sentence.

It requires verifying `intent`, `signer` and `isGuest`, so it now says those comparisons are exact: a value or property name differing only in case must be refused rather than case-folded.

## Checked and deliberately left unchanged

- **ADR-49 (Signed Fetch V2)** — already correct. It concatenates the metadata verbatim (`'x-identity-metadata:' + Metadata`), and its lowercasing applies to header *names* and `Content-Type`, which is correct HTTP canonicalization rather than metadata folding. It is also status Draft.
- **ADR-42, ADR-72** — matched a `lowercase` search but are unrelated (Ethereum address formats; an HTTP method check in a service worker).
- **ADR-183, ADR-208** — mention signed fetch but do not restate the payload construction.

## How to Test

```sh
make build
```

Passes, and ADR-44 / ADR-289 / ADR-180 render with the new sections.

### DIFF
--- a/content/ADR-180-communication-adapters.md
+++ b/content/ADR-180-communication-adapters.md
@@ -81,7 +81,7 @@ Here is the list of officially supported protocols
 
 The signed login exists as part of a "handshake" mechanism to enable authenticated log-ins and access control to different communication services like Livekit.
 
-The verifications of the [signed fetch](/adr/ADR-44) must look for the following properties: `intent`, `signer`, and `isGuest`.
+The verifications of the [signed fetch](/adr/ADR-44) must look for the following properties: `intent`, `signer`, and `isGuest`. These properties are bound to the signature exactly as delivered, so they must be compared exactly: a value or property name differing only in case must be refused rather than case-folded, which would otherwise let a re-spelled property read as absent.
 
 ```typescript
 // connectionString = "signed-login:https://my-server/authenticate-comms"
--- a/content/ADR-289-explorer-client-scenes-signed-fetch.md
+++ b/content/ADR-289-explorer-client-scenes-signed-fetch.md
@@ -44,7 +44,7 @@ This specification impacts:
 
 ### Overview
 
-The Explorer Client Signed Fetch functionality extends [ADR-44](/adr/ADR-44) by including scene-specific metadata in the signature. This metadata is included in the `X-Identity-Metadata` header and is part of the signed payload, ensuring its integrity and authenticity.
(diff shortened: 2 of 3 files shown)


nachomazzara (commented):
LGTM

PR #324, opened 2026-09-15
ADR-318: Audio playback position reports

LautaroPetaccio:
## Summary

Proposes extending `PBAudioEvent` so renderers report the clip playback position while an `AudioSource` plays, in addition to the media state changes they already report. A scene asks for the reports with a new `PBAudioSource.report_playback_position` flag; state changes are reported whatever the flag says.

## Why

Scenes that align gameplay or visuals with sound have no way to learn where a clip's playhead is: `PBAudioSource.current_time` is a write-only seek, and `PBAudioEvent` carries only a media state and a counter. Measured on the Unity explorer, audible playback starts 100 to 250 ms after the scene asks for it, with a different value on each start, which is as wide as a rhythm game's hit window. `PBVideoEvent` already reports `current_offset` and `tick_number` for video.

## Contents

- Problem statement with the measurements
- Protocol fields, and why the reports are opt-in rather than always on
- Why a report carries a tick rather than a timestamp, and why the per-tick clock history belongs in the client library rather than in every scene
- What the reports do not remove: the renderer's sampling granularity, and the output latency between the decoder position a report carries and the moment a sample leaves the speaker
- Renderer behaviour, including bounds on the reporting rate
- Alternatives considered (scheduled playback, readable `current_time`, higher tick rates, onset detection), compatibility

## Notes on two decisions

**The reporting rate is bounded here rather than left to each renderer.** A survey of the three explorers found they already disagree on video: one reports on every frame the position changes, one at most once per second of playback, one once the position has moved half a second. Scenes cannot write against that. The document now states at most one report per tick, since the tick is the resolution of the stamp and finer carries nothing a scene can place in time, and at least one per second of playback so a scene converges promptly. All three existing implementations sit inside those bounds.

**The specification stays protocol-only.** An earlier revision showed the SDK implementation and a scene usage example. Both are gone; what an SDK does with these fields is a client-library concern and the code lives in the js-sdk-toolchain pull request. What survives is the requirement that the tick-to-clock correlation belongs in the library rather than in each scene, and that a report must not be dropped when its tick is missing from the history. That requirement is also the protocol rationale for stamping a tick rather than a wall-clock time.

Status: Draft. Numbering: `main` reaches 316 and open PRs use 317, so 318 is the next free number; a few open PRs sit out of sequence at 323, 400 and 401.

## Related

- Protocol change: https://github.com/decentraland/protocol/pull/488
- SDK change: https://github.com/decentraland/js-sdk-toolchain/pull/1624
- Unity explorer change: https://github.com/decentraland/unity-explorer/pull/10123

## Related pull requests

- Test scene: https://github.com/decentraland/sdk7-test-scenes/pull/99
- SDK skills: https://github.com/decentraland/sdk-skills/pull/97


pravusjif:
The only potential problem with that is that both the scene and the engine could modify the property (`current_time`) "at the same time", and also if any other property of the component is modified from the scene, the engine receives the full component data through the CRDT command, so wouldn't know if the `current_time` was modified or not by the scene, since that property was designed to tell the engine to START at a specific time of the audio.

What we can do is follow the same approach we do for videos, [constantly updating their videoEvent with the current time in that other component](https://github.com/decentraland/unity-explorer/blob/fe6974465b0d2e3a70eeb1ba2da3cb87df27e654/Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs#L63), WDYT?

LautaroPetaccio:
You're right on both counts, and that is how it works now: the position goes in `PBAudioEvent`, not in `PBAudioSource`. The renderer never writes `PBAudioSource`, so `current_time` stays a write-only seek, neither side clobbers the other, and a full-component put can never be mistaken for a new seek. It is the `VideoEventsSystem` pattern applied to audio (decentraland/unity-explorer#10123).

The abstract names the `PBVideoEvent` precedent, and the rejected "readable `current_time`" alternative carries both of your reasons.

content/ADR-318-audio-playback-position-reports.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Renderer behaviour
+
+- Renderers keep appending an `AudioEvent` on every media state change, as today.
+- While an `AudioSource` is in `MS_PLAYING`, renderers also append a report at least every 15 scene ticks (about twice a second at the reference tick rate), carrying `tick_number`, `current_offset` and `clip_length`. This is the same mechanism the Unity explorer uses for `PBVideoEvent` in [`VideoEventsSystem`](https://github.com/decentraland/unity-explorer/blob/fe6974465b0d2e3a70eeb1ba2da3cb87df27e654/Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs#L63), applied to audio sources.
```
mikhail-dcl: >renderers also append a report at least every 15 scene ticks (about twice a second at the reference tick rate)

It'd be a very strange approach leading to heavy desync.
`VideoEventsSystem` doesn't do it - it writes the component off every frame if the video has progressed

content/ADR-318-audio-playback-position-reports.md
```diff
@@ -0,0 +1,106 @@
...
+`registerAudioEventsEntity` keeps its current semantics and only fires on state changes, so existing scenes receive no extra callbacks from the periodic reports.
+
+### Scene usage
+
+A scene that keeps its own song clock records that clock against `EngineInfo.tickNumber` each frame, keeping a short history so a report's tick can be looked up after the fact, then aligns on every report:
+
+```ts
+audioEventsSystem.registerAudioPlaybackEntity(drums, report => {
+  if (report.currentOffset === undefined || report.tickNumber === undefined) return
+  const heardAt = songTimeAtTick(report.tickNumber)          // ms, from EngineInfo.tickNumber samples
```
mikhail-dcl: Please present the implementation of the function

mikhail-dcl: After all the accuracy depends on this function as it must evaluate (somehow) the round trip of the CRDT

content/ADR-318-audio-playback-position-reports.md
```diff
@@ -0,0 +1,106 @@
...
+`audioEventsSystem` in `@dcl/ecs` gains:
+
+- `registerAudioPlaybackEntity(entity, callback)` and `removeAudioPlaybackEntity(entity)`: the callback runs for every report, position updates included.
+- `getAudioPlayback(entity)`: the latest report that carries `current_offset`, or `undefined`.
+
+`registerAudioEventsEntity` keeps its current semantics and only fires on state changes, so existing scenes receive no extra callbacks from the periodic reports.
+
+### Scene usage
+
+A scene that keeps its own song clock records that clock against `EngineInfo.tickNumber` each frame, keeping a short history so a report's tick can be looked up after the fact, then aligns on every report:
```
mikhail-dcl: So it's here described but without the implementation - I believe it should be part of SDK - otherwise each scene will try to come up with its own history of snapshots?

mikhail-dcl (changes requested):

content/ADR-318-audio-playback-position-reports.md
LautaroPetaccio: You're right. There is no cadence now: a report goes out whenever the state or the clip position differs from the last propagated value, exactly as `VideoEventsSystem` does. A playing clip reports every frame, and a paused or stopped one emits nothing.

content/ADR-318-audio-playback-position-reports.md
LautaroPetaccio: You're right, and that is how it works now: the history is in the SDK, not in each scene. `audioEventsSystem` keeps the scene clock per `EngineInfo.tickNumber`, and `registerAudioPlaybackEntity` hands the callback `{ report, sceneTime, offset }` with `sceneTime` already the clock at the tick the renderer sampled in. `getSceneTimeAtTick` exposes the same lookup on its own, for raw reports and for video. It is the only registration, so there is no rawer, shorter-named alternative for a scene to reach for instead.

content/ADR-318-audio-playback-position-reports.md
LautaroPetaccio: It is in the document now, under the SDK section. A per-frame system accumulates the engine's delta time and stores it under the current `EngineInfo.tickNumber`, keeping about 128 ticks, and `getSceneTimeAtTick` is a map lookup.

LautaroPetaccio: You're right that the accuracy sits here. The document now states what the report does and does not remove, instead of implying the result is exact.

What it removes is the transit time: the report names the tick it was sampled in, and the scene reads its own clock for that tick, so however long the report spent in the queue does not enter the result.

What it leaves is two smaller terms. The playhead is read once per renderer frame and advances in audio-buffer steps, so the reading is quantised. And `current_offset` is the decoder's read position, not the moment a sample leaves the speaker: the mixer buffer, the driver and the device add tens of milliseconds on top, in one direction and roughly constant per device, and no field carries it. That residue is a per-session constant for the scene to calibrate, and a renderer able to estimate its own output latency should expose it.

LautaroPetaccio: Correcting myself here: you were right that this is where the accuracy lives, and my previous answer overstated it.

The tick stamp does remove the transport delay, and that part stands. The report says which tick it was sampled in, the scene looks up its own clock for that tick, so however many ticks the report spent in the queue, it does not enter the comparison. What was wrong is the sentence after it, that the remaining error is then one tick.

Two terms survive. The first is sampling granularity: `current_offset` is read once per renderer frame, from a playhead that advances in audio-buffer steps, so the reading is quantised to the coarser of the two. The second is the one I argued away and should not have: what a renderer exposes is the decoder's read position, not the moment the sample leaves the speaker. The mixer buffer, the driver and the device add tens of milliseconds on top of it, in the same direction and roughly constant for a given machine and output device. That delay is invisible to this report, and it is precisely the delay a scene syncing to what it can actually hear wants removed.

So the honest claim is narrower than the one I made: the report removes the transport delay and leaves a per-session constant that the scene has to calibrate. I have rewritten that section (574fca8) to say so, and noted that a renderer able to estimate its own output latency should expose it, since that is the version of this which would really close the gap rather than bound it.

One detail worth recording because it bears on the history you asked about: a renderer publishes `EngineInfo.tick_number` and the reports sampled in that tick in the same batch, so the lookup usually resolves inside the frame that receives the report. The history still has to tolerate being asked for a tick it has not recorded, and the SDK side must not drop a report when that happens. I am fixing that ordering on the SDK PR.

LautaroPetaccio:
Status across the five pull requests.

**Protocol** (decentraland/protocol#488): three optional fields on `PBAudioEvent`, `tick_number`, `current_offset` and `clip_length`.

**Renderer** (decentraland/unity-explorer#10123): a report goes out whenever the media state changed or the clip position moved, which is the rule `VideoEventsSystem` applies to video. The emit rule is `ShouldReport`, beside `IsNaturalFinish`, covered by tests that need no audio device. Protocol bindings are regenerated, so the branch compiles as it stands.

**SDK** (decentraland/js-sdk-toolchain#1624): `registerAudioPlaybackEntity` delivers the newest position report of the frame, already resolved against the scene clock at the tick it was sampled in. `getSceneTimeAtTick` exposes the same lookup for scenes handling raw reports and for video.

**Test scene** (decentraland/sdk7-test-scenes#99): three beat cubes contrasting the scene's own clock, the resolved reading, and the same reading timed on arrival, against a clip that beeps on every whole second.

**Documentation and skills**: decentraland/documentation#609 and decentraland/sdk-skills#97.

One behaviour worth flagging: a finished clip reports a zero offset rather than `clip_length`, because players rewind the playhead on stop and at the end of a clip. A scene tracking completion reads the state transition out of playing.

The SDK and renderer pull requests both hold a branch pin on `@dcl/protocol`, so their dependency checks stay red until #488 merges.

mikhail-dcl (approved):

LautaroPetaccio:
You're right, and the spec is protocol-only now. The SDK implementation and the scene usage example are gone.

What I kept is the requirement rather than the code: the tick-to-clock correlation belongs in the client library instead of each scene, and a report must not be dropped when its tick is missing from the history. That one is really protocol rationale, since it is why a report carries a tick instead of a wall-clock timestamp, so removing it would leave a future reader asking why the field is shaped that way. The implementation itself is in decentraland/js-sdk-toolchain#1624.

I also moved the accuracy limits out of that section. Sampling granularity and output latency bound what any renderer can deliver through these fields, so they belong next to the protocol rather than under one client library.

